PR #27, opened 2025-05-15
EVM Wasmd Query

jhelison:
# Description

This creates a new Wasmd contract query:
- Add `eth_call`
- Add timeout and other fee requirements

## Type of change

# How Has This Been Tested?

New tests added


wasmbinding/evm/queries.go
```diff
@@ -0,0 +1,128 @@
...
+	to := common.HexToAddress(call.Contract)
+	data, err := hexutil.Decode(call.Data)
+	if err != nil {
+		return nil, err
+	}
+
+	// Build the arguments
+	args := evmtypes.TransactionArgs{
+		To:   &to,
+		Data: (*hexutil.Bytes)(&data),
```
Thaleszh, line 73: This looks crazy

wasmbinding/evm/queries.go
```diff
@@ -0,0 +1,128 @@
...
+	bz, err := json.Marshal(&args)
+	if err != nil {
+		return nil, err
+	}
+
+	// Build the request
+	req := evmtypes.EthCallRequest{
+		Args:            bz,
+		GasCap:          emvconfig.DefaultGasCap,
+		ChainId:         int64(qp.evmKeeper.GetParams(ctx).ChainConfig.ChainId),
```
Thaleszh: Isn't it more readable to set a variable with the chain id and just convert here? I quite dislike the nesting

wasmbinding/evm/queries_test.go
```diff
@@ -0,0 +1,102 @@
...
+// TestHandleEVMQuery tests the HandleEVMQuery function
+func TestHandleEVMQuery(t *testing.T) {
+	// Setup the app
+	actor := helpers.RandomAccountAddress()
+	app, ctx := helpers.SetupCustomApp(t, actor)
+
+	// Deploy the counter contract
+	contractAddr := deployCounter(t, ctx, app)
+	// Increment the counter
+	incrementCounter(t, ctx, app, contractAddr)
```
Thaleszh, line 24: Isn't the increment unecessary here?

Thaleszh (approved):

wasmbinding/evm/queries.go
```diff
@@ -0,0 +1,128 @@
...
+		ProposerAddress: sdk.ConsAddress(ctx.BlockHeader().ProposerAddress),
+	}
+
+	// Build a timeout and wrap the context
+	timeout := emvconfig.DefaultEVMTimeout
+	var cancel context.CancelFunc
+	var ctxWrapped context.Context
+	ctxWrapped, cancel = context.WithTimeout(ctx, timeout)
```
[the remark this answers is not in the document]
jhelison: No, you are wrong

wasmbinding/evm/queries_test.go
jhelison: Its needed, I feel weird checking for zero since it would be a default response. So to make the testing clear we check for 1 after the increment

wasmbinding/evm/queries.go
jhelison: Yeah, the deal here is:
- data is already a `[]byte` and `hexutil.Byte` is also a `[]byte`. So `(*hexutil.Bytes)(&data)` is a type assertion, since they are 'related' we don't get the normal `ok` as we have in normal type assertions.
- Now for the whole pointer thing. The data is already a reference, so we take its value with `&data`. The `Data` from the `evmtypes.TransactionArgs` needs to be a reference that's why `hexutil.Bytes` has the `*`

wasmbinding/evm/queries.go
jhelison: Sure, addressed

PR #36, opened 2025-06-12
App integration with Oracle and Gasless fees

jhelison:
# Description

This makes the app integration and implements the gasless fees.
This also applies a correct implementation for Begin and Endblocker

## Type of change

# How Has This Been Tested?

Tests are passing

### DIFF
--- a/ante/feeless.go
+++ b/ante/feeless.go
@@ -2,6 +2,7 @@ package ante
 
 import (
 	"errors"
+	"math"
 
 	"cosmossdk.io/collections"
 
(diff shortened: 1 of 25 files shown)


app/apptesting/test_suite.go
```diff
@@ -278,3 +279,17 @@ func GenerateTestAddrs() (string, string) {
 	invalidAddr := sdk.AccAddress("invalid").String()
 	return validAddr, invalidAddr
 }
+
+// we need to make this deterministic (same every test run), as content might affect gas costs
```
Thaleszh, line 283: This is a TODO right? It's just the way it's phrased looked like an explanation of what the function is doing

Thaleszh: I don't get why the account info could change gas costs tho

x/oracle/abci.go
```diff
@@ -10,10 +10,9 @@ import (
 	"github.com/kiichain/kiichain/v1/x/oracle/utils"
 )
 
-// MidBlocker is the function executed when each block has been completed
-// this function get the votes from the validators, calculate the exchange rate using
+// MidBlocker get the votes from the validators, calculate the exchange rate using
```
Thaleszh: Still using old name?

x/oracle/utils/period.go
```diff
@@ -3,12 +3,9 @@ package utils
 import sdk "github.com/cosmos/cosmos-sdk/types"
 
 const (
-	BlocksPerMinute = uint64(75)
+	BlocksPerMinute = uint64(17)
```
Thaleszh, line 6: Isn't it weird to set up block time as a hard variable? Why is this needed

jhelison: Just to initialize the params, it can be any arbitrary value. So we the least to get it right.
This is similar to the staking window:
https://github.com/cosmos/cosmos-sdk/blob/e77156858f14e2725144cd81c5e39c7872db6d3c/x/slashing/types/params.go#L12

x/oracle/abci.go
jhelison: Damn auto rename not working, solved

app/apptesting/test_suite.go
jhelison: This was the same comment as they left from the EVM module, it relates to how bech32 is calculated that it consume gas

Thaleszh (approved):

ante/feeless.go
```diff
@@ -62,8 +64,8 @@ func (gd GaslessDecorator) IsTxGasless(ctx sdk.Context, tx sdk.Tx) (bool, error)
 	for _, msg := range tx.GetMsgs() {
 		switch m := msg.(type) {
 		case *oracletypes.MsgAggregateExchangeRateVote:
-			// Check if the message message is a gasless message
-			return gd.MsgAggregateExchangeRateVoteIsGasless(ctx, m)
+			// Check if the message message is a feeless message
+			return gd.MsgAggregateExchangeRateVoteIsFeeless(ctx, m)
 		default:
 			// We can return that its not gasless
```
Thaleszh, line 70: gasless remark

jhelison: Addressed

PR #38, opened 2025-06-13
E2E tests for reward

Thaleszh:
# Description

Adds E2E testing for rewards module

## Type of change

- [ ] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

make test


.github/codecov.yml
```diff
@@ -22,6 +22,7 @@ ignore:
   - "**/*.pb.gw.go"
   - "**/*.md"
   - "*.rst"
+  - "**/cli/*"
```
jhelison: This feels weird. I think that I would rather evaluate the CLI tests case by case.
```suggestion
```

x/rewards/client/cli/tx.go
```diff
@@ -0,0 +1,117 @@
...
+
+			msg := types.NewMsgChangeSchedule(clientCtx.GetFromAddress().String(), schedule)
+
+			return tx.GenerateOrBroadcastTxCLI(clientCtx, cmd.Flags(), msg)
+		},
+	}
+
+	flags.AddTxFlagsToCmd(cmd)
+	return cmd
+}
```
jhelison: Functions missing docstrings

x/rewards/types/reward.go
```diff
@@ -42,6 +42,8 @@ func CalculateReward(blockTime time.Time, schedule ReleaseSchedule) (sdk.Coin, e
 	// Truncate to int, it will be a coin amt after all
 	amountToRelease := math.LegacyNewDecFromInt(remaining.Amount).Mul(releaseProportion).TruncateInt()
 
+	// Make sure at least one coin will be distributed if there is still a distribution to happen
+	amountToRelease = math.MaxInt(amountToRelease, math.NewInt(1))
```
jhelison, line 46: This is weird, can you explain it?

tests/e2e/e2e_test.go
jhelison, line 1: You should add the module to the tests on`tests/e2e/e2e_rest_regression_test.go`

tests/e2e/e2e_rewards_test.go
```diff
@@ -0,0 +1,226 @@
...
+	s.Require().True(schedule.ReleasedAmount.Amount.LT(finalSchedule.ReleasedAmount.Amount))
+	s.Require().True(schedule.Active)
+
+	// Check validator balance change
+	finalRewardsB, err := queryRewardFrom(chainEndpoint, validatorB.String(), valOperAddressB)
+	s.Require().NoError(err)
+	initialAkii := initialRewardsB.Rewards.AmountOf(denom)
+	finalAkii := finalRewardsB.Rewards.AmountOf(denom)
+	s.T().Logf("Balance amt before %s vs after %s", initialAkii.String(), finalAkii.String())
+	s.Require().True(initialAkii.LT(finalAkii))
```
jhelison: Balance change isn't enough. The TXs have fees that mess with your tests.
Maybe you should do a schedule of 30 seconds, check at 15 if rewards increase and check at 30+n to see if all the expected schedule was released under the time.
The rewards should be `current rewards > initial rewards + scheduled rewards`

x/rewards/types/reward.go
Thaleszh: Yeah, this is just a small trick to make sure I am releasing something each block instead of 0, this maintaining the releaser active.

We are flagging the release schedule as inactive if it doesn't have anything to distribute, but there is an issue where the reward distribution is rounded down to 0 on calculations (i.e result would be 0.5 and truncated to 0).
This is to solve that specific scenario, making sure at least 1 akii is distributed if there a remaining amt to distribute. If there is nothing to distribute, the reward calculation will still return zero. We can meet to talk over, but I couldn't think of another solution that wouldn't change the behavior a lot.

tests/e2e/e2e_rewards_test.go
Thaleszh: So I am checking specificly for the rewards info, the 'balance amt' might be misleading here. Exactly since there could be other rewards, I can't make sure the calculation is exact in the middle parts because the execution of the code itself is not deterministic, since we are awaiting commands. So it is very hard to pick up a mid point and figuring how much was distributed and making sure that it is actually that on the balances. That's why I was just checking an increase there. There is also the fact there are multiple validators on the chain and the reward will be split among them.

After the whole distribution has already happened, we can refine that amt but still can't be sure the validator didn't get an extra reward, so it's hard to use exact numbers here. I will add a change to check at the end of the schedule that it received a bunch of it.

.github/codecov.yml
Thaleszh: Added only the rewards one, then.

x/rewards/client/cli/tx.go
Thaleszh: addressed

tests/e2e/e2e_test.go
Thaleszh: Oh yeah, forgot about this. Addressed

PR #47, opened 2025-06-23
Oracle E2E

jhelison:
# Description

Add simple e2e tests to the oracle module:
- Test feeless TXs
- Test voting
- Test feeder assignment
- Test Slashing params

## Type of change

- [ ] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

This is a test


tests/e2e/e2e_oracle_test.go
```diff
@@ -0,0 +1,340 @@
...
+	s.Require().NoError(err, "failed to get balance for %s after voting", voterAddr.String())
+	s.Require().Equal(balance.Amount, newBalance.Amount, "balance should remain the same after fee-less vote")
+
+	// If we vote again, the balance should change
+	s.execAggregateVote(s.chainA, 0, "1000akii", validatorAddress, voterAddr.String(), kiichainHomePath, Fee.String(), nil)
+
+	// Get the new balance after the second vote
+	newBalanceAfterSecondVote, err := getSpecificBalance(chainEndpoint, voterAddr.String(), akiiDenom)
+	s.Require().NoError(err, "failed to get balance for %s after second vote", voterAddr.String())
+	s.Require().NotEqual(newBalance.Amount, newBalanceAfterSecondVote.Amount, "balance should change after second vote")
```
Thaleszh: Why not equal? Shouldn't it be less due to fee?

tests/e2e/e2e_oracle_test.go
```diff
@@ -0,0 +1,340 @@
...
+	validatorAddress := sdk.ValAddress(voterAddr).String()
+
+	// Get the penalty counter for the validator
+	queryPenaltyCounter, err := queryPenaltyCounter(chainEndpoint, validatorAddress)
+	s.Require().NoError(err, "failed to query penalty counter for validator %s", validatorAddress)
+
+	// Both the success and the abstain votes should be above zero
+	fmt.Println(queryPenaltyCounter)
+	s.Require().Greater(queryPenaltyCounter.VotePenaltyCounter.SuccessCount, uint64(0), "success penalty counter should be greater than zero")
+	s.Require().Greater(queryPenaltyCounter.VotePenaltyCounter.AbstainCount, uint64(0), "abstain penalty counter should be greater than zero")
```
Thaleszh: I don't get how the penalty should happen here when nothing is done except querying

tests/e2e/e2e_oracle_test.go
```diff
@@ -0,0 +1,340 @@
...
+		5*time.Second,
+	)
+
+	// Get the current height
+	currentHeight := s.getLatestBlockHeight(c, 0)
+	next := ((currentHeight/BlocksPerPeriod)+1)*BlocksPerPeriod + 1
+
+	// Wait for the height, but log first with the current height
+	s.T().Logf("Waiting for the next block height %d on chain %s. Current block %d", next, c.id, currentHeight)
+	s.waitUntilPassedHeight(s.chainA, 0, next)
```
Thaleszh, line 199: No checks after passed height?

tests/e2e/e2e_oracle_test.go
jhelison: Interestingly that what SEI does too. Slashing is extremely hard to check and mess with other tests lined

tests/e2e/e2e_oracle_test.go
jhelison: Addressed, now it checks for less than

tests/e2e/e2e_oracle_test.go
jhelison: No, we just swap the epoch for the module. The first test actually validates it by voting twice on the same epoch.

tests/e2e/e2e_oracle_test.go
jhelison: Interestingly, that's what SEI checks because this test can be damaging to other running tests. Also, it's quite complicated to validate. So in this case we count if the penalty counters are being updated

Thaleszh (approved):

PR #59, opened 2025-07-05
test: add TestBasicCoinMath for sdk.Int operation

senjaparades:
### 📌 PR Description

Add a unit test `TestBasicCoinMath` for simple `sdk.Int` addition in `x/bank/keeper`.  
This test ensures the accuracy of basic math using Cosmos SDK types.

### ✅ Changes
- Add `keeper_test.go` with a simple test using `require.Equal`

### 🧪 How to Test
```bash
go test -v ./x/bank/keeper/

### DIFF
--- a/x/bank/keeper/keeper_test.go
+++ b/x/bank/keeper/keeper_test.go
@@ -1,0 +1,17 @@
+package keeper
+
+import (
+    "testing"
+
+    sdk "cosmossdk.io/math"
+    "github.com/stretchr/testify/require"
+)
+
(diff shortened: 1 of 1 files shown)


Thaleszh:
Hello there,
This test verifies basic behavior of cosmos SDK math and is unrelated to Kiichain itself. If it was placed anywhere, it would be in cosmos' math library, which already has such tests.
The second thing is that it's placed in a module we did not modify from cosmos SDK. We don't have a `/x/bank` in this repo and creating a folder like that would just produce confusion.

PR #95, opened 2025-08-06
feat: add simple dApp UI

Muladi12:
Added a basic dApp interface under `app/simple-dapp` including wallet connect and basic token interaction.

Please review and advise if anything else should be added.


Muladi12:
Changelog updated ✅
This PR is ready for review 🚀

jhelison:
@Muladi12 
This PR isn't the place for a sample Daap. It also copies the whole blockchain implementation and moves it to another folder

PR #99, opened 2025-08-07
Add basic whitelist check to feeless transaction logic

justmortysmith:
Hi team,

While going through the `ante/feeless.go` file, I noticed that any address can potentially submit feeless transactions. This could expose the chain to unwanted spam or abuse, especially if not combined with other safeguards.

I added a basic whitelist mechanism that allows only selected module or system accounts (e.g., `gov`, `distribution`) to send feeless transactions. Other addresses will proceed through the normal fee deduction logic.

Here’s what I changed:
- Added a small helper function to check whether the sender is on the feeless whitelist.
- Limited feeless handling to only those whitelisted addresses.
- Kept the list static/hardcoded to keep things simple for now.

Let me know if you'd prefer to make the whitelist configurable via params or governance later on; happy to help with that too.

Thanks for the awesome work on KiiChain!


jhelison:
@justmortysmith 
Your PR states that any arbitrary address can execute the antehandler. But that's incorrect:
- Although all addressed will go through the antehandler, only messages of type `oracletypes.MsgAggregateExchangeRateVote` can go through the fee less
  - See on https://github.com/KiiChain/kiichain/blob/main/ante/feeless.go#L63-L73
- Even for messages of this type, it has to be from a bondended validator who hasn't voted for the epoch
  - See on https://github.com/KiiChain/kiichain/blob/main/ante/feeless.go#L81-L111

So I don't see the point where you state that all addresses are susceptible to fee-less transactions

Also, looking at your implementation, I don't see code related to the whitelisting of addresses.

Even though your typo fixes on comments and variable renaming are valid (but not related to the PR description), it breaks the chain as you can see on pipelines:
- [E2E / test-e2e (pull_request)](https://github.com/KiiChain/kiichain/actions/runs/16811167707/job/47617105568?pr=99)
- [Lint / golangci-lint (pull_request)](https://github.com/KiiChain/kiichain/actions/runs/16811167730/job/47617106067?pr=99)
- [Liveness check / liveness-test (pull_request)](https://github.com/KiiChain/kiichain/actions/runs/16811167711/job/47617105789?pr=99)
- [Test / repo-analysis (pull_request)](https://github.com/KiiChain/kiichain/actions/runs/16811167719/job/47617397762?pr=99)
- [Test / tests (pull_request)](https://github.com/KiiChain/kiichain/actions/runs/16811167719/job/47617105850?pr=99)

jhelison:
Closing the PR due to inactivity

PR #114, opened 2025-09-15
fix(oracle): add denom regex validation and remove fallback in keeper

Ethaga:
# Description
This PR introduces stricter denom validation and fixes misleading defaults in keeper.

- Added regex validation for denom in `types.go` (`[a-z][a-z0-9/]{2,64}`) to ensure valid input format.
- Updated keeper implementation to return error when denom does not exist, instead of falling back to a default value.

This change improves data integrity and prevents potential misleading results when querying non-existent denoms.

## Type of change

# How Has This Been Tested?
- [x] Added unit tests for invalid/empty denom.
- [x] Verified queries with non-existent denom now return proper error instead of default value.

### DIFF
--- a/precompiles/ibc/tx.go
+++ b/precompiles/ibc/tx.go
@@ -12,10 +12,32 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 )
 
-// Transfer does a IBC transfer with custom timeout options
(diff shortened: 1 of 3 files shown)


jhelison:
@Ethaga, Pipelines are failing, can you check?

PR #137, opened 2025-10-08
Fix reentrance problems on the Wasmd precompile

jhelison:
# Description

This applies fixes and tests related to the three main issues on the Wasmd precompile as reported on https://github.com/KiiChain/kiichain/blob/main/contrib/docs/wasmd_precompile_vulnerability.md

## Reetrance protection

Reetrance protection is ensured by using a transient key under the EVM db state:
- The transient lock, is applied for all senders (signers) and for each nonce
- This blocks re-calls from the same signer and nonce multiple times under the same TX execution
- Since a transient key is used, the same is cleared before the block is commit

Tests:
- Tests were updated to cover reentrance attacks

## Log

A log was printing full messages on `execute`:
- There was a possibility of nodes having issues with the message while logging
- To fix the log was updated to not carry the message 

Tests:
- Tests were updated or created to ensure logging is now constrained

## Gas

Tests:
Tests were updated to ensure gas usage

## Type of change

# How Has This Been Tested?

Tests added or updated

### DIFF
--- a/precompiles/wasmd/realworld_attack_test.go
+++ b/precompiles/wasmd/realworld_attack_test.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/stretchr/testify/require"
 
 	cmn "github.com/cosmos/evm/precompiles/common"
 	"github.com/cosmos/evm/precompiles/testutil"
@@ -60,8 +61,8 @@ func (s *WasmdPrecompileTestSuite) TestRealWorldReentrancyAttack() {
 		attackAttempts := 5
 		successfulReentries := 0
 
+		stateDB := s.GetStateDB()
 		for attempt := 1; attempt <= attackAttempts; attempt++ {
-			stateDB := s.GetStateDB()
 			contract, ctx := testutil.NewPrecompileContract(
 				s.T(),
 				s.Ctx,
@@ -140,6 +141,7 @@ func (s *WasmdPrecompileTestSuite) TestRealWorldReentrancyAttack() {
 			s.T().Log("  🔒 Funds irrecoverable once stolen")
 			s.T().Log("  👥 All users affected")
 			s.T().Log("")
+			require.Fail(s.T(), "REENTRANCY VULNERABILITY EXISTS")
 		} else if successfulReentries == 1 {
 			s.T().Log("")
 			s.T().Log("✅ Only one execution succeeded - Guard may be present")
--- a/precompiles/wasmd/realworld_attack_test.go
+++ b/precompiles/wasmd/realworld_attack_test.go
@@ -1,18 +1,23 @@
 package wasmd_test
 
 import (
-	"encoding/hex"
 	"encoding/json"
 
-	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/stretchr/testify/require"
 
 	storetypes "cosmossdk.io/store/types"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
 	cmn "github.com/cosmos/evm/precompiles/common"
 	"github.com/cosmos/evm/precompiles/testutil"
 
 	wasmdprecompile "github.com/kiichain/kiichain/v5/precompiles/wasmd"
 )
+
+// Note:
+// The test TestRealWorldDataLeakage was removed because its causer was removed
+// The test TestCompareRealWorldVsUnitTest was removed because it was redundant
 
 // TestRealWorldReentrancyAttack simulates a real-world attack scenario
 // This test is closer to what would happen on an actual running chain
@@ -36,8 +41,10 @@ func (s *WasmdPrecompileTestSuite) TestRealWorldReentrancyAttack() {
 		// Simulate funding the victim contract (like a DeFi pool)
 		initialFunds := sdk.NewCoins(sdk.NewInt64Coin("ukii", 10000000))
 		victimAddr, _ := sdk.AccAddressFromBech32(victimContract)
-		s.App.BankKeeper.MintCoins(s.Ctx, "evm", initialFunds)
-		s.App.BankKeeper.SendCoinsFromModuleToAccount(s.Ctx, "evm", victimAddr, initialFunds)
+		err := s.App.BankKeeper.MintCoins(s.Ctx, "evm", initialFunds)
+		require.NoError(s.T(), err)
+		err = s.App.BankKeeper.SendCoinsFromModuleToAccount(s.Ctx, "evm", victimAddr, initialFunds)
+		require.NoError(s.T(), err)
 
 		victimBalance := s.App.BankKeeper.GetBalance(s.Ctx, victimAddr, "ukii")
 		s.T().Logf("✓ Victim contract funded: %s", victimBalance.String())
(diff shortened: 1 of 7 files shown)


jhelison:
@coderabbitai review

precompiles/wasmd/tx.go
```diff
@@ -106,3 +116,49 @@ func (p Precompile) Execute(
 	// Return the response
 	return method.Outputs.Pack(true)
 }
+
+// ensureLock ensures that a reentrancy lock is set and not broken for the target contract
+// Reentrance lock is built using: precompile address, origin and origin nonce
+//   - Args are avoid to build the lock key, since the attacker may manipulate it
```
Thaleszh:
```suggestion
//   - Args are avoided to build the lock key, since the attacker may manipulate it
```

precompiles/wasmd/tx.go
```diff
@@ -106,3 +116,49 @@ func (p Precompile) Execute(
...
+//
```
Thaleszh, line 123:
```suggestion
```

precompiles/wasmd/tx.go
```diff
@@ -106,3 +116,49 @@ func (p Precompile) Execute(
...
+func (p Precompile) ensureLock(
+	origin common.Address,
+	stateDB vm.StateDB,
+	method *abi.Method,
+) error {
+	// Build the lock key
+	lockKey := buildReentrancyLockKey(p.Address(), origin, stateDB.GetNonce(origin))
+
+	// Check if already locked
+	if stateDB.GetTransientState(p.Address(), lockKey) == common.BytesToHash([]byte{1}) {
```
Thaleszh, line 135: I don't fully get the `common.BytesToHash([]byte{1})`
The return of the function is a hashed data? And you're just checking if it's true?

precompiles/wasmd/tx.go
```diff
@@ -106,3 +116,49 @@ func (p Precompile) Execute(
...
+		)
+	}
+
+	// Set lock
+	stateDB.SetTransientState(p.Address(), lockKey, common.BytesToHash([]byte{1}))
+	return nil
+}
+
+// buildReentrancyLockKey builds a deterministic lock key:
+//
```
Thaleszh, line 148:
```suggestion
```

precompiles/wasmd/tx.go
```diff
@@ -106,3 +116,49 @@ func (p Precompile) Execute(
...
+	// Check if already locked
+	if stateDB.GetTransientState(p.Address(), lockKey) == common.BytesToHash([]byte{1}) {
+		return fmt.Errorf(
+			"reentrancy detected in precompile %s, method %s",
+			p.Address().Hex(), method.Name,
+		)
+	}
+
+	// Set lock
+	stateDB.SetTransientState(p.Address(), lockKey, common.BytesToHash([]byte{1}))
```
Thaleszh, line 143: So this is working almost like a temporary log? I don't see this information getting deleted / freed. 
You mentioned it was cleared after each block since is transient, how does this work?

precompiles/wasmd/security_test.go
```diff
@@ -0,0 +1,389 @@
...
+			// But the real issue is that they get logged with raw binary
+			if tc.expectError {
+				s.Require().Error(err, tc.description)
+			}
+		})
+	}
+}
+
+// TestGasHandling tests if gas limits are properly enforced
+// Bug Report Issue #2: Missing Gas Handling
```
Thaleszh, line 159: I see this bug report test here, but I don't understand what could have fixed it.

precompiles/wasmd/tx.go
```diff
@@ -78,13 +83,18 @@ func (p Precompile) Execute(
...
+	}
+
 	// Log the call
 	p.Logger(ctx).Debug(
 		"tx called",
 		"method", method.Name,
 		"args", fmt.Sprintf(
-			"{ contract: %s, msg: %s, sender: %s }",
-			msg.Contract, msg.Msg, msg.Sender,
+			"{ contract: %s, sender: %s }",
```
Thaleszh, line 96: Nice catch from whom reported the bug

precompiles/wasmd/security_test.go
```diff
@@ -0,0 +1,389 @@
...
+	storetypes "cosmossdk.io/store/types"
+
+	cmn "github.com/cosmos/evm/precompiles/common"
+	"github.com/cosmos/evm/precompiles/testutil"
+
+	wasmdprecompile "github.com/kiichain/kiichain/v5/precompiles/wasmd"
+)
+
+// TestInvalidCodeIDValidation tests if the precompile properly validates code_id
+// Bug Report Issue #4: No Input Validation
```
Thaleszh, line 18: I see this bug report test here, but I don't understand what in the PR could have fixed it.

precompiles/wasmd/realworld_attack_test.go
```diff
@@ -0,0 +1,254 @@
...
+
+	cmn "github.com/cosmos/evm/precompiles/common"
+	"github.com/cosmos/evm/precompiles/testutil"
+
+	wasmdprecompile "github.com/kiichain/kiichain/v5/precompiles/wasmd"
+)
+
+// Note:
+// The test TestRealWorldDataLeakage was removed because its causer was removed
+// The test TestCompareRealWorldVsUnitTest was removed because it was redundant
```
Thaleszh, line 20: Irrelevant text for our context?

precompiles/wasmd/tx.go
```diff
@@ -106,3 +116,49 @@ func (p Precompile) Execute(
...
+// buildReentrancyLockKey builds a deterministic lock key:
+//
+//	H( "wasmd.precompile.reentrancy.lock:", precompileAddr, originAddr, originNonce[8] )
+func buildReentrancyLockKey(
+	precompileAddr common.Address,
+	origin common.Address,
+	originNonce uint64,
+) common.Hash {
+	var nonceBytes [8]byte
+	binary.BigEndian.PutUint64(nonceBytes[:], originNonce)
```
Thaleszh, line 156: I don't understand what this part is doing. 

Also why the `originNonce[8]` in the comment

precompiles/wasmd/realworld_attack_test.go
jhelison: These were the tests removed from the original test suite.
We can leave it here for the auditors

precompiles/wasmd/security_test.go
jhelison: The bug was never there, the validations happen during the message execution on the Wasm layer.
This just makes sure that the bug is resolved

precompiles/wasmd/security_test.go
jhelison: Nothing fixed, but the test was updated to ensure that all is fine

precompiles/wasmd/tx.go
jhelison: Yep

precompiles/wasmd/tx.go
jhelison: No, above we set as 1 to avoid reetrance, so here we check as 1 again

precompiles/wasmd/tx.go
jhelison: This is just bits conversion nonsense.

Just uint64 serialization to build the key

precompiles/wasmd/tx.go
jhelison: A transient state is just an in-memory state handling where we never commit the changes at the end of the block.

On Cosmos-SDK, we also have the transient keys, check here:
- https://github.com/KiiChain/kiichain/blob/main/app/keepers/keys.go#L74-L80

These keys are never `removed`, they are just not committed. Since on each block, we start a new db, the transient store is always empty.

precompiles/wasmd/tx.go
Thaleszh: Found it, thanks

precompiles/wasmd/tx.go
Thaleszh: This is awesome, nice!

precompiles/wasmd/security_test.go
Thaleszh: Hmmmm, so the test that failed due to gas at Webisoft was not working properly?

jhelison: So, its seems that they have assumed that the function `testutil.NewPrecompileContract` was generating a gas aware context. But its not:
- It actually tests the gas post-execution, this is not what we want [reference](https://github.com/cosmos/evm/blob/main/precompiles/testutil/testing.go#L17)
- So the changes I applied the gas limit directly on the context: https://github.com/KiiChain/kiichain/blob/fix/wasmd-precom/precompiles/wasmd/realworld_attack_test.go#L220-L221

Thaleszh (approved):

AndresRamirez9912 (approved):

PR #158, opened 2025-10-31
Bump EVM cosmos version to v0.4.2

Thaleszh:
# Description
- Bump EVM cosmos version to v0.4.2
- Pushes some changes on EVM side that weren't mirrored in the codebase

## Type of change

# How Has This Been Tested?
Unit tests and E2E tests


ante/ante_cosmos.go
```diff
@@ -55,13 +55,18 @@ func NewCosmosAnteHandler(options HandlerOptions) sdk.AnteHandler {
 	// Skip the feemarket decorator is needed
 	if UseFeeMarketDecorator {
 		// This wraps using the gasless decorator
+		var txFeeChecker ante.TxFeeChecker
+		if options.DynamicFeeChecker {
+			txFeeChecker = evmante.NewDynamicFeeChecker(options.FeeMarketKeeper)
+		}
```
jhelison, line 61: Why was this needed?

ante/tx_listener.go
```diff
@@ -0,0 +1,35 @@
...
+
+type PendingTxListener func(common.Hash)
+
+type TxListenerDecorator struct {
+	pendingTxListener PendingTxListener
+}
+
+// NewTxListenerDecorator creates a new TxListenerDecorator with the provided PendingTxListener.
+// CONTRACT: must be put at the last of the chained decorators
+func NewTxListenerDecorator(pendingTxListener PendingTxListener) TxListenerDecorator {
```
jhelison: Why this was migration to our side? I can imagine that this comes from a lib correct? Maybe we can reuse from source by importing

tests/e2e/e2e_rate_limit_test.go
```diff
@@ -260,6 +260,7 @@ func (s *IntegrationTestSuite) testUpdateRateLimit() {
 	submitGovFlags := []string{configFile(proposalUpdateRateLimitAtomFilename)}
 	depositGovFlags := []string{strconv.Itoa(proposalCounter), depositAmount.String()}
 	voteGovFlags := []string{strconv.Itoa(proposalCounter), "yes"}
+	time.Sleep(5 * time.Second)
```
jhelison: Why this sleep?

Thaleszh: So, the error that was happening was that the TX sequence was incorrect. This was due it not processing the previous TX in time for this one. I am not sure how to do it without a simple sleep

ante/ante_cosmos.go
Thaleszh: It wasn't, but it was a change they added to their own repo. In the v0.5.0 it will be needed since it will need the context.

ante/tx_listener.go
Thaleszh: It wasn't, I didn't see the import when I checked it, but it's there. Pushed a fix

jhelison (dismissed):

jhelison (approved):

PR #164, opened 2025-11-07
Remove native fallback price from fee abstraction

Thaleszh:
# Description

Remove native fallback price from fee abstraction and short-circuits pre block when there is not price for kii

## Type of change

# How Has This Been Tested?
`make test-unit`


x/feeabstraction/keeper/oracle.go
```diff
@@ -37,7 +37,13 @@ func (k Keeper) CalculateFeeTokenPrices(ctx sdk.Context) error {
...
 	if !ok {
-		baseTokenPrice = params.FallbackNativePrice
+		// Disable fee abstraction if there is no pricing
+		k.Logger(ctx).Debug("akii has no price, feeabstraction disabled")
+		params.Enabled = false
+		if err = k.Params.Set(ctx, params); err != nil {
+			return err
+		}
+		return nil
 	}
```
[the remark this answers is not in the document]
Thaleszh: Yeah, this might be annoying

x/feeabstraction/ante/cosmos/fee_test.go
```diff
@@ -53,6 +53,10 @@ func TestDeductFeeDecorator(t *testing.T) {
 	// Create the funder account
 	app.AccountKeeper.SetAccount(ctx, app.AccountKeeper.NewAccountWithAddress(ctx, founder))
 
+	// Reactivate keeper since it was disabled on abci
+	err := app.FeeAbstractionKeeper.Params.Set(ctx, types.DefaultParams())
+	require.NoError(t, err)
```
jhelison, line 58: This is weird, on `SetupWithContext` we should be setting the module with defaults. So why re-set here?

x/feeabstraction/ante/evm/mono_decorator_test.go
```diff
@@ -110,6 +110,10 @@ func TestMonoDecorator(t *testing.T) {
 	err := app.FeeMarketKeeper.SetParams(ctx, feeMarketParams)
 	require.NoError(t, err)
 
+	// Reactivate keeper since it was disabled on abci
+	err = app.FeeAbstractionKeeper.Params.Set(ctx, types.DefaultParams())
+	require.NoError(t, err)
```
jhelison, line 115: Same as https://github.com/KiiChain/kiichain/pull/164/files#r2515152069. Why set this as the default here?

The module auto-disables itself on setup?

x/feeabstraction/keeper/abci.go
```diff
@@ -44,15 +44,17 @@ func (k Keeper) WriteFeeTokenPricesMetrics(ctx context.Context) error {
...
+		if token.Enabled {
+			if floatPrice, err := token.Price.Float64(); err == nil {
+				telemetry.ModuleSetGauge(
+					types.ModuleName,
+					float32(floatPrice),
+					"fee_token_price",
+					token.Denom,
+				)
+			}
 		}
```
jhelison, line 61: Small nitpick

```suggestion
	for _, token := range feeTokenPrices.Items {
	    // Check if the token is enabled
	    if !token.Enabled {
	        continue  
	    }
	    
		// Set a module metric for each enabled token
		if floatPrice, err := token.Price.Float64(); err == nil {
			telemetry.ModuleSetGauge(
				types.ModuleName,
				float32(floatPrice),
				"fee_token_price",
				token.Denom,
			)
}
```

x/feeabstraction/keeper/fee_test.go
```diff
@@ -28,6 +28,10 @@ func (s *KeeperTestSuite) TestConvertNativeFee() {
 	// Default erc20 address to use in tests
 	DefaultFirstERC20 := "0x80b5a32E4F032B2a058b4F29EC95EEfEEB87aDcd"
 
+	// Reactivate keeper since it was disabled on abci
+	err := s.app.FeeAbstractionKeeper.Params.Set(s.ctx, types.DefaultParams())
+	s.Require().NoError(err)
```
jhelison, line 33: Related to the other comments with the enable with Default Params

x/feeabstraction/keeper/oracle.go
```diff
@@ -37,7 +37,10 @@ func (k Keeper) CalculateFeeTokenPrices(ctx sdk.Context) error {
 	// Find the price for the base token
 	baseTokenPrice, ok := twapPriceMap[params.NativeOracleDenom]
 	if !ok {
-		baseTokenPrice = params.FallbackNativePrice
+		// Disable fee abstraction if there is no pricing
+		k.Logger(ctx).Debug("akii has no price, feeabstraction disabled")
```
jhelison: Use the param when logging:

```suggestion
		k.Logger(ctx).Debug(fmt.Printf("%s has no price, feeabstraction disabled", params.NativeOracleDenom))
```

x/feeabstraction/ante/cosmos/fee_test.go
Thaleszh: What happens is that there is a validation that the pre-blockers and all block operations are working before moving on. This makes the first iteration run, see no TWAP, and disable the module all together. 
All these tests do something that checks if the module is enabled before continuing, and were failing due to that.

x/feeabstraction/keeper/abci.go
Thaleszh: Nice, applied

x/feeabstraction/keeper/oracle.go
Thaleszh: Addressed

jhelison (approved):

jhelison (approved):

PR #174, opened 2025-11-25
Add EVM mempool

Thaleszh:
# Description

Plugs in EVM mempool

## Type of change

# How Has This Been Tested?

Locally plus all tests

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -42,6 +42,7 @@ import (
 	servertypes "github.com/cosmos/cosmos-sdk/server/types"
 	"github.com/cosmos/cosmos-sdk/testutil/testdata"
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkmempool "github.com/cosmos/cosmos-sdk/types/mempool"
 	"github.com/cosmos/cosmos-sdk/types/module"
 	"github.com/cosmos/cosmos-sdk/types/msgservice"
 	sigtypes "github.com/cosmos/cosmos-sdk/types/tx/signing"
@@ -60,8 +61,10 @@ import (
 	// EVM
 	"github.com/cosmos/evm/ante"
 	evmencoding "github.com/cosmos/evm/encoding"
+	evmmempool "github.com/cosmos/evm/mempool"
 	srvflags "github.com/cosmos/evm/server/flags"
 	cosmosevmtypes "github.com/cosmos/evm/types"
+	evmtypes "github.com/cosmos/evm/x/vm/types"
 
 	kiiante "github.com/kiichain/kiichain/v6/ante"
 	"github.com/kiichain/kiichain/v6/app/keepers"
@@ -100,6 +103,7 @@ type KiichainApp struct { //nolint: revive
 	// EVM v0.4.1
 	clientCtx          client.Context
 	pendingTxListeners []ante.PendingTxListener // from "github.com/cosmos/evm/ante"
+	EVMMempool         *evmmempool.ExperimentalEVMMempool
 
 	// the module manager
 	mm           *module.Manager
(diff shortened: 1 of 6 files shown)


tests/e2e/e2e_rate_limit_test.go
```diff
@@ -374,7 +374,5 @@ func (s *IntegrationTestSuite) testIBCTransfer(expToFail bool) {
 		s.Require().NotNil(res.RateLimit)
 		s.Require().Equal(sdkmath.NewInt(0), res.RateLimit.Flow.Inflow)
 		s.Require().NotEqual(sdkmath.NewInt(0), res.RateLimit.Flow.Outflow)
-	} else {
-		time.Sleep(5 * time.Second)
 	}
```
GiovaniGuizzo, line 379: Why remove this?

GiovaniGuizzo (dismissed):

tests/e2e/e2e_rate_limit_test.go
Thaleszh: This was a sleep we added a while ago due to the nounces not matching, exactly because of no mempool.

Pretty much the last line of this function sends a function to exec, and the code that runs it execute another similar function soon after, without waiting for this one to check. This made the second TX fail because it had a higher nounce than expected

app/app.go
```diff
@@ -273,6 +277,32 @@ func NewKiichainApp(
 	maxGasWanted := cast.ToUint64(appOpts.Get(srvflags.EVMMaxTxGasWanted))
 	app.setAnteHandler(app.txConfig, maxGasWanted, appOpts)
 
+	// set the EVM priority nonce mempool
+	if evmtypes.GetChainConfig() != nil {
+		// TODO: Get the actual block gas limit from consensus parameters
```
GiovaniGuizzo, line 282: What about this TODO?

tests/e2e/e2e_evm_test.go
```diff
@@ -103,6 +103,50 @@ func (s *IntegrationTestSuite) testEVM(jsonRPC string) {
...
+	s.Run("Testing out of order nounce", func() {
+		// Send 3 Txs with nonce
+		// Nonce +0
+		tx, err := EVMSendWithNonce(client, evmAccount.key, evmAccount.address, s.chainB.evmAccount.address, amount, nil, nonce)
+		s.Require().NoError(err)
+
+		// Nonce +2
+		tx2, err := EVMSendWithNonce(client, evmAccount.key, evmAccount.address, s.chainB.evmAccount.address, amount, nil, nonce+2)
+		s.Require().NoError(err)
+
```
GiovaniGuizzo, line 135: I would add here a sleep of like 500ms to make sure the nonce + 2 was indeed submitted BEFORE nonce + 1. If all 3 are submitted at the same time, it might still work without mempool.

tests/e2e/e2e_evm_test.go
```diff
@@ -103,6 +103,50 @@ func (s *IntegrationTestSuite) testEVM(jsonRPC string) {
...
+		// Wait for successful first tx
+		s.waitForTransaction(client, tx, evmAccount.address)
+
+		// Wait for successful second tx
+		s.waitForTransaction(client, tx2, evmAccount.address)
+
+		// Wait for successful third tx
+		s.waitForTransaction(client, tx3, evmAccount.address)
```
GiovaniGuizzo, line 147: Don't you have to check the receipt for status === "confirmed"?

app/app.go
Thaleszh: This was taken directly from the EVM code, they have this TODO there. It is an option to have this hard value on the params, but we don't have it

tests/e2e/e2e_evm_test.go
Thaleszh: Inside the wait transaction I check the status to be sucessful. We only have 0x0 (failed) and 0x1 (successful)

tests/e2e/e2e_evm_test.go
Thaleszh: Added

GiovaniGuizzo (approved):

PR #180, opened 2025-12-24
Bump EVM to v0.5.1

Thaleszh:
# Description

Bump EVM to v0.5.1

## Type of change

# How Has This Been Tested?


ante/ante_evm.go
```diff
@@ -9,15 +9,21 @@ import (
 )
 
 // newMonoEVMAnteHandler creates the sdk.AnteHandler implementation for the EVM transactions
-func newMonoEVMAnteHandler(options HandlerOptions) sdk.AnteHandler {
-	return sdk.ChainAnteDecorators(
+func newMonoEVMAnteHandler(ctx sdk.Context, options HandlerOptions) sdk.AnteHandler {
```
jhelison, line 12: The changes in this function are weird.
I don't know about the specifics, but this may cause weird initialization issues:
- If the newMonoEVMAnteHandler is initialized only during the app initialization, this would mean that the values broadcasted as params would not reflect changes during runtime (only during app initialization)
- This may cause consensus issues

This is correct or the way I'm thinking about the initialization is incorrect?

cmd/kiichaind/cmd/root.go
```diff
@@ -451,6 +413,12 @@ func (a appCreator) newApp(
 		baseapp.SetIAVLCacheSize(cast.ToInt(appOpts.Get(server.FlagIAVLCacheSize))),
 	}
 
+	viperAppOpts, ok := appOpts.(*viper.Viper)
+	if !ok {
+		panic("appOpts is not viper.Viper")
+	}
+	viperAppOpts.Set(srvflags.EVMChainID, kiichain.KiichainID)
```
jhelison: We are setting the flag for EVM chain ID directly as `kiichain.KiichainID`?
What happens if I pass the flag to the application, it would still be replaced by `kiichain.KiichainID`?

Maybe we should check any value exists before we set it

app/keepers/keepers.go
```diff
@@ -402,8 +413,9 @@ func NewAppKeeper(
 		appKeepers.FeeMarketKeeper,
 		&appKeepers.ConsensusParamsKeeper,
 		&appKeepers.Erc20Keeper,
+		evmChainID,
 		tracer,
-	)
+	).WithDefaultEvmCoinInfo(CoinInfo)
```
jhelison, line 418: This is supposed to be hardcoded?

app/app.go
```diff
@@ -213,6 +210,7 @@ func NewKiichainApp(
 	app.mm.SetOrderPreBlockers(
 		upgradetypes.ModuleName,
 		authtypes.ModuleName,
+		evmtypes.ModuleName,
```
jhelison, line 213: This is new, do know why this is was needed?

precompiles/oracle/oracle.go
```diff
@@ -122,17 +112,8 @@ func (p Precompile) run(evm *vm.EVM, contract *vm.Contract, readOnly bool) (bz [
 		// If default error out
 		return nil, fmt.Errorf(cmn.ErrUnknownMethod, method.Name)
 	}
-	if err != nil {
-		return nil, err
-	}
-
-	// Check the gas cost
-	cost := ctx.GasMeter().GasConsumed() - initialGas
```
jhelison, line 130: Weird that the gas cost section was removed, maybe it was embedded on the execution process for pre-compiles?

x/feeabstraction/ante/evm/account_verification.go
```diff
@@ -28,7 +28,7 @@ func VerifyIfAccountExists(
 	from common.Address,
 ) error {
 	// Only EOA are allowed to send transactions.
-	if account != nil && account.IsContract() {
+	if account != nil && account.HasCodeHash() {
```
jhelison, line 31: Now its codeHash? Same function at the end of the day correct?

ante/ante_evm.go
Thaleszh: Whenever a proposal is passed to change params, doesn't it restart the node? If not, it does sound like it could be an issue.

This was both in their EVMD and in their migration guide. Maybe they are preparing to decouple this information from params?

I am not sure what drove them to receive just the params. The feemarket one is only checking if the base fee is enabled and how much it is, for the remainder it still uses the keeper.

The evm params is checked for message validation.

cmd/kiichaind/cmd/root.go
Thaleszh: That is true, it's a good check to have. 

When I made this I was thinking of our case of keeping current way, using the chain ID as source of truth and ignoring the rest.

I just checked where the Kiichain ID is being used, it's here and on another custom app config on the root cmd

precompiles/oracle/oracle.go
Thaleszh: Yes, a lot of the boiler code was moved inside the precompile type.

x/feeabstraction/ante/evm/account_verification.go
Thaleszh: Yeah, it checks if there is any contract code on the address

app/app.go
Thaleszh: Explanation I got was from a dark commit message:
```
however, we're getting a recovered panic on *restarting* the chain due to prepareproposal running before preblocker:

+0x120\ngithub.com/cosmos/cosmos-sdk/baseapp.(*BaseApp).PrepareProposal.func1()\n\tgithub.com/cosmos/cosmos-sdk@v0.53.4/baseapp/abci.go:433 +0x38\npanic({0x1061c8160?, 0x10912faa0?})\n\truntime/panic.go:783 +0x120\ngithub.com/cosmos/evm/x/vm/types.GetEVMCoinDecimals(...)\n\tgithub.com/cosmos/evm@v0.2.0/x/vm/types/denom_config.go:61\ngithub.com/cosmos/evm/x/vm/types.ConvertAmountTo18DecimalsLegacy(...)\n\tgithub.com/cosmos/evm@v0.2.0/x/vm/types/scaling.go:17\ngithub.com/cosmos/evm/x/vm/wrappers.FeeMarketWrapper.GetBaseFee({{_, _}}, {{0x10698a810, 0x109225540}, {0x1069b51c0, 0x14001f4c200}, {{0x0, 0x0}, {0x16dc0f689, 0x4}, ...}, ...})\n\tgithub.com/cosmos/evm@v0.2.0/x/vm/wrappers/feemarket.go:38 +0x68\ngithub.com/cosmos/evm/x/vm/keeper.Keeper.GetBaseFee({{0x1069b4890, 0x14001eddeb0}, {0x10695e0b0, 0x14003560450}, {0x10695e240, 0x14003560490}, 0x140035637a0, {0x1400189d3e0, 0x14, 0x20}, ...}, ...)\n\tgithub.com/cosmos/evm@v0.2.0/x/vm/keeper/keeper.go:355 +0xe4\ngithub.com/cosmos/evm/mempool.(*ExperimentalEVMMempool).getIterators(0x14003cbd180, {0x10698a848?, 0x14002c39c08?}, {0x109225540, 0x0, 0x0})\n\tgithub.com/cosmos/evm@v0.2.0/mempool/mempool.go:439 +0x20c\ngithub.com/cosmos/evm/mempool.(*ExperimentalEVMMempool).SelectBy(0x14003cbd180, {0x10698a848?, 0x14002c39c08?}, {0x109225540?, 0x14000085ce8?, 0x104cce034?}, 0x140044cd540)\n\tgithub.com/cosmos/evm@v0.2.0/mempool/mempool.go:370 +0xe0\ngithub.com/cosmos/cosmos-sdk/types/mempool.SelectBy({0x10698a848?, 0x14002c39c08?}, {0x10698b108?, 0x14003cbd180?}, {0x109225540?, 0x0?, 0x0?}, 0x140044cd540)\n\tgithub.com/cosmos/cosmos-sdk@v0.53.4/types/mempool/mempool.go:58 +0xa0\ngithub.com/cosmos/evm/evmd.NewExampleApp.(*DefaultProposalHandler).PrepareProposalHandler.func3({{0x10698a810, 0x109225540}, {0x1069b51c0, 0x14001f4c200}, {{0x0, 0x0}, {0x16dc0f689, 0x4}, 0xd, {0x3507e5b0, ...}, ...}, ...}, ...)\n\tgithub.com/cosmos/cosmos-sdk@v0.53.4/baseapp/abci_utils.go:294 +0x250\ngithub.com/cosmos/cosmos-sdk/baseapp.(*BaseApp).PrepareProposal(0x14002d32488, 0x140015f42c0)\n\tgithub.com/cosmos/cosmos-sdk@v0.53.4/baseapp/abci.go:439
```

Thaleszh: There is also this:
- [\#816](https://github.com/cosmos/evm/pull/816) Avoid nil pointer when RPC requests execute before evmCoinInfo initialization in PreBlock with defaultEvmCoinInfo fallback.

Looks like it's related to the defaultEVMCoinInfo setup.

app/keepers/keepers.go
Thaleszh: Yeah, it's a fix done by the Mantra team that got implemented here. When I first started doing this it wasn't there.

- [\#790](https://github.com/cosmos/evm/pull/790) fix panic in historical query due to missing EvmCoinInfo.
- [\#816](https://github.com/cosmos/evm/pull/816) Avoid nil pointer when RPC requests execute before evmCoinInfo initialization in PreBlock with defaultEvmCoinInfo fallback.

jhelison (dismissed):

jhelison (approved):

PR #196, opened 2026-01-02
feat: add evm upgrade v5

Thaleszh:
# Description

Adds local upgrade to v7 with EVM changes

## Type of change

# How Has This Been Tested?
local upgrade script


app/upgrades/v7_0/upgrade.go
```diff
@@ -0,0 +1,93 @@
...
+				Denom:    params.DisplayDenom,
+				Exponent: 18,
+				Aliases:  nil,
+			},
+		},
+		Base:    params.BaseDenom,
+		Display: params.DisplayDenom,
+		Name:    "Kii",
+		Symbol:  "KII",
+		URI:     "example_uri",
```
jhelison: Should this example be here? Maybe would be better to remove the two params?

jhelison: URI and URIHash

Thaleszh: Will remove them, yep. I left them just to check if we had anything to add here.

jhelison (approved):

PR #213, opened 2026-01-07
tokenfactory: avoid panics during genesis initialization

raddenpattah:
Summary
This PR replaces panic calls in tokenfactory genesis initialization
with proper error logging and graceful handling.

Motivation
Panics during genesis processing can crash the chain
due to malformed or unexpected genesis data.

Changes
- Removed `panic(err)` calls
- Added structured error logging
- Continued or safely returned instead of hard crashing

Impact
- No business logic changes
- Improved chain stability
- Safer for operators and test environments

Testing
- Ran `go test -tags=test ./x/tokenfactory/...`
- All relevant tests pass; one authz test is skipped by design (existing TODO)

### DIFF
--- a/x/tokenfactory/keeper/genesis.go
+++ b/x/tokenfactory/keeper/genesis.go
@@ -12,22 +12,38 @@ func (k Keeper) InitGenesis(ctx sdk.Context, genState types.GenesisState) {
 	if genState.Params.DenomCreationFee == nil {
 		genState.Params.DenomCreationFee = sdk.NewCoins()
 	}
+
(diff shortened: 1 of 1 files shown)


jhelison:
@raddenpattah 

Thanks for the contribution. The chain must fail if the genesis doesn't get loaded correctly.
You can see evidence about the usage of Panic on genesis initialization on Cosmos-SDK:
- https://github.com/cosmos/cosmos-sdk/blob/main/x/bank/keeper/genesis.go

Due to this, this PR is being closed

Issue #224, opened 2026-01-08
[SAFETY][SDK v0.53] Custom modules lack RegisterInvariants - Missing automatic state corruption checks


raddenpattah:
## Bug Description
Kiichain's custom modules (`x/oracle`, `x/rewards`, `x/tokenfactory`) do not implement the `RegisterInvariants` method required by the `AppModule` interface in Cosmos SDK v0.53+. This disables the runtime's automatic invariant checking system, removing a critical fail-safe mechanism against state corruption.

**Technical Context:** The legacy `x/crisis` module is deprecated. The new architecture requires each module to implement `RegisterInvariants` for automatic safety checks.

## Steps to Reproduce
**1. Verify inconsistent `RegisterInvariants` implementation across modules :**
```bash
# x/oracle has it (partial implementation)
grep -n "RegisterInvariants" ./x/oracle/module.go
# Output: 140,141 - shows method exists

# x/rewards and x/tokenfactory lack it
grep -n "RegisterInvariants" ./x/rewards/module.go ./x/tokenfactory/module.go
# Output: (empty) - confirming the safety gap
```
**2. Confirm modules use the modern AppModule interface :**
```bash
grep -l "AppModule interface" ./x/oracle/module.go ./x/rewards/module.go ./x/tokenfactory/module.go
# All three modules use the interface
```
## Expected Behavior

- Each `AppModule` implements `RegisterInvariants` with module-specific checks
- Runtime automatically verifies invariants at configured intervals
- Chain halts immediately if any invariant breaks (fail-fast safety)

## Actual Behavior

- `x/rewards` and `x/tokenfactory` have no `RegisterInvariants` implementations
- `x/oracle` has an implementation (proof-of-concept added) but other modules lack it
- No consistent automatic state consistency checking across custom modules
- Chain continues operating even with corrupted internal state in affected modules

## Environment
Required information:

- Kiichain version / commit: `github.com/kiichain/kiichain/v7` (from `go.mod`)
- Network: All networks (local/testnet/mainnet) - source code structural issue
- OS: Ubuntu 25.10 (Linux, but issue is OS-independent)
- Node configuration: N/A (issue is in module source implementation)
- SDK Version: `v0.53.5-0.20251030204916-768cb210885c`
- Go Version: Go 1.23.8 (meets SDK v0.53's Go 1.21+ requirement)
- Affected Modules: `x/rewards` (fully), `x/tokenfactory` (fully), `x/oracle` (has partial implementation)

[...]

PR #226, opened 2026-01-12
[DRAFT] POC: custom guard pattern for rewards module

raddenpattah:
## Purpose

This PR introduces a proof-of-concept (POC) for a custom guard pattern in the rewards module, following the discussion in issue #224.
The goal is to explore application-controlled, off-consensus state consistency checks in the SDK v0.53+ era.
This is not a proposal to introduce new consensus invariants or to modify block execution semantics.

## Context

In issue #224, @jelison mentioned that Kiichain will follow Cosmos-SDK paradigms, while still allowing redesigns or the addition of custom guards at the module level.

This PR implements that invitation as a concrete and scoped example in the rewards module.

## Important Scope & Execution Clarification

These guards are not executed automatically.

Specifically:

- They are not registered via `RegisterInvariants`
- They are not wired into `BeginBlock`, `EndBlock`, or `ProcessProposal`
- They are not consensus-critical and cannot halt the chain

The guards are implemented and exposed only.
Execution is explicit and application-controlled.

Intended use cases include:

- Integration tests
- Migrations
- Debugging
- Operational monitoring

## What’s Implemented
Three reward-related guards are provided:

1. Schedule Guard
Ensures released amount does not exceed total amount.

2. Community Pool Guard
Ensures no negative balances exist in the community pool.

3. Params Guard
Validates reward-related denom configuration.

## Design Pattern

- A minimal keeper interface is defined to avoid circular dependencies and to keep guard logic read-only.
- Each guard is implemented as a pure, deterministic function that inspects state and reports violations without mutating state.
- An aggregator provides a single entry point to collect all rewards-related guards, making them easy to invoke from tests or tooling.

## Characteristics

- Read-only
- Off-consensus
- Optional and application-controlled execution
- Purely additive (no API changes)
- Covered by unit tests (using -tags=test)

## Intent & Discussion Points

**This PR is marked as DRAFT and is intended for architectural discussion, not immediate merge.**

Feedback is welcome on:

- Whether this guard-based pattern fits Kiichain’s long-term direction
- Boundary placement between guards, SDK invariants, and tests
- When (if ever) such guards should be promoted to stronger guarantees
- Maintenance overhead versus safety benefits

## Collaboration Note

This PR is exploratory by design.

If the pattern proves useful, it can be refined or extended to other modules.
If not, it still serves as a concrete reference for future safety and tooling discussions.

### DIFF
--- a/x/rewards/keeper/invariants.go
+++ b/x/rewards/keeper/invariants.go
@@ -1,0 +1,29 @@
+package keeper
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/kiichain/kiichain/v7/x/rewards/keeper/invariants"
+	"github.com/kiichain/kiichain/v7/x/rewards/types"
+)
+
+// Implement KeeperInterface untuk invariants
+func (k Keeper) ReleaseScheduleGet(ctx sdk.Context) (types.ReleaseSchedule, error) {
+	return k.ReleaseSchedule.Get(ctx)
+}
+
+func (k Keeper) RewardPoolGet(ctx sdk.Context) (types.RewardPool, error) {
+	return k.RewardPool.Get(ctx)
+}
+
+func (k Keeper) ParamsGet(ctx sdk.Context) (types.Params, error) {
+	return k.Params.Get(ctx)
+}
+
+// GetAllInvariants returns all invariant checkers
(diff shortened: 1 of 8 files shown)


jhelison:
PR converted to draft as the title suggests

PR #232, opened 2026-01-14
fix(oracle): handle errors in pickReferenceDenom instead of panic

### LINKED ISSUE #230: [BUG] Ignored error in TotalBondedTokens breaks oracle threshold
## Bug Description

in `x/oracle/tally.go` line 22, error from `TotalBondedTokens()` ignored with `_`. if call fails, totalBondedTokens = 0, thresholdVotes = 0. with 0 threshold any ballot passes.

funny thing is Params.Get error right below is handled but this one just ignored.

## Steps to Reproduce

1. check code at `x/oracle/tally.go` line 22
2. see error ignored with `_`
3. if TotalBondedTokens fails, threshold becomes 0

```bash
grep -n "TotalBondedTokens" x/oracle/tally.go
# 22:    totalBondedTokens, _ := k.StakingKeeper.TotalBondedTokens(ctx)
```

## Expected Behavior

should handle error like other calls in same function. return error or log it.

## Actual Behavior

error silently ignored:

```go
totalBondedTokens, _ := k.StakingKeeper.TotalBondedTokens(ctx)  // error ignored
```

if error:
- totalBondedTokens = 0
- totalBondedPower = 0  
- thresholdVotes = 0
- any ballot passes

## Environment

- Kiichain version / commit: v6.1.0 / `79bdc96`
- Network: testnet
- File: `x/oracle/tally.go` line 22

## Impact Assessment

- single validator could manipulate oracle prices
- threshold bypass
- Security risk: medium

Reference: https://github.com/KiiChain/kiichain/blob/v6.1.0/x/oracle/tally.go#L22

## Reporter Declaration

- [x] This report is NOT generated by AI
- [x] I personally tested and verified this issue

ngapaxs:
# fix(oracle): handle errors in pickReferenceDenom instead of panic

- Handle error from TotalBondedTokens instead of ignoring with `_`
- Return error from Params.Get instead of panic
- Update pickReferenceDenom signature to return error
- Update caller in abci.go to handle the error

Fixes #230

## Description

In `x/oracle/tally.go`, the `pickReferenceDenom` function has two error handling issues:

1. Line 22: Error from `TotalBondedTokens()` is ignored with `_`
2. Line 28: Error from `Params.Get()` causes panic instead of returning error

This PR restructures `pickReferenceDenom` to return an error to the oracle EndBlocker, as suggested by maintainer @Thaleszh.

## Type of change

## How Has This Been Tested?

- [x] Code review - verified error handling logic
- [x] Checked that function signature change propagates correctly to caller in abci.go

### DIFF
--- a/x/oracle/tally_test.go
+++ b/x/oracle/tally_test.go
@@ -1,6 +1,8 @@
 package oracle
 
 import (
+	"context"
+	"fmt"
 	"testing"
 
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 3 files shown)


Thaleszh (changes requested):
Hello there! Nice PR ongoing, but there are a few things requiring change:

- The code you pushed is using a different spacing lint. Run `make lint-fix` to ensure it has the same pattern as the repo's. It will also make it much better for reviewing. You can check if it's working by reviewing the change tab on the PR, it should only highlight your changes.
- Each change must be documented in the changelog. Please add the fix you are pushing under `Unreleased > Fixed`
- It is also good practice to add tests cases on `tally_test.go` that checks the new behaviors introduced.

x/oracle/tally_test.go
```diff
@@ -220,3 +222,40 @@ func TestPickReferenceDenomError(t *testing.T) {
...
+
+func (m MockStakingKeeperError) TotalBondedTokens(ctx context.Context) (math.Int, error) {
+	return math.Int{}, fmt.Errorf("mock error: failed to get total bonded tokens")
+}
+
+func (m MockStakingKeeperError) PowerReduction(ctx context.Context) math.Int {
+	return sdk.DefaultPowerReduction
+}
+
+func TestPickReferenceDenomTotalBondedTokensError(t *testing.T) {
```
Thaleszh, line 242: These three functions are missing docstring describing their behavior

Thaleszh (changes requested):
Looking excellent. Just have a nitpick on missing docstring on some tests

Thaleszh (approved):
Looking good, waiting for pipelines and then will merge

PR #233, opened 2026-01-14
Fix swagger generation

Thaleszh:
# Description

Fixes faulty swagger generation by:
- Renaming conflicting IDs on config.json
- Renaming conflicting refs on temporary files
- Running `make proto-swagger-gen` (might require sudo depending on docker permissioning)

## Type of change

# How Has This Been Tested?

Manually, locally.

# PR Checklist:

Make sure each step was done:


proto/scripts/swagger-protoc.sh
```diff
@@ -1,13 +1,14 @@
-#!/usr/bin/env bash
-
-set -eo pipefail
+#!/bin/sh
+# set -e
```
Thaleszh, line 2: Do you know why this in specific could make the makefile fail? Its running inside a docker container as well

Makefile
```diff
@@ -387,6 +387,14 @@ proto-swagger-gen:
...
 	@$(protoImage) sh ./proto/scripts/swagger-protoc.sh
 
+	@echo "Fixing definition conflicts"
+	@./proto/scripts/rename-conflicting-defs.sh
+
+	@echo "Fixing Id conflicts"
+	@./proto/scripts/fix-conflicting-ids.sh
+
+	@$(protoImage) sh ./proto/scripts/swagger-combine.sh
+
```
Thaleszh, line 397: The file that creates the dependency might have a different permission than the docker container, which may lead to the temporary files not being deleted.

Do you know how I could fix it?

x/feeabstraction/types/query.pb.go
```diff
@@ -339,7 +339,6 @@ func _Query_FeeTokens_Handler(srv interface{}, ctx context.Context, dec func(int
 	return interceptor(ctx, in, info, handler)
 }
 
-var Query_serviceDesc = _Query_serviceDesc
```
jhelison, line 342: Do you know why these changes in the proto-generated files happened?

proto/buf.yaml
```diff
@@ -8,7 +8,7 @@ name: buf.build/kiichain/kiichain
 deps:
   - buf.build/cosmos/gogo-proto
   - buf.build/cosmos/cosmos-proto
-  - buf.build/cosmos/cosmos-sdk:v0.47.0
+  - buf.build/cosmos/cosmos-sdk:v0.53.5
```
jhelison, line 11: Interesting. Could swear that this was not available before. But good thing it is available now

proto/scripts/swagger-protoc.sh
jhelison: Can you explain more? I was going to respond about the `-e`.

proto/scripts/swagger-protoc.sh
```diff
@@ -1,13 +1,14 @@
...
 
 mkdir -p ./tmp-swagger-gen
+
 proto_dirs=$(find ./proto ./tmp_deps -path -prune -o -name '*.proto' -print0 | xargs -0 -n1 dirname | sort | uniq)
+
 for dir in $proto_dirs; do
   # generate swagger files (filter query files)
   query_file=$(find "${dir}" -maxdepth 1 \( -name 'query.proto' -o -name 'service.proto' \))
-  if [[ ! -z "$query_file" ]]; then
+  if [ -n "$query_file" ]; then
```
jhelison, line 11: What happened here?

Makefile
jhelison: You can change the permission of the generated folders using Docker:
```
chmod -R 755 /path/to/folder
```
But it seems that all generated fine on my side.

proto/scripts/rename-conflicting-defs.sh
```diff
@@ -0,0 +1,66 @@
...
+      ' "$file" > "${file}.tmp"
+
+      mv "${file}.tmp" "$file"
+      modified=1
+    fi
+  done
+
+  if [ "$modified" = "0" ]; then
+    :  # No-op command (do nothing)
+  fi
```
jhelison: What is this?

proto/scripts/swagger-protoc.sh
Thaleszh: When I try to run the makefile command, it fails if the `set -e` is there. But it doesn't fail for all the other scripts.

Makefile
Thaleszh: Thats good to know

proto/scripts/swagger-protoc.sh
Thaleszh: I switched the whole file from bash to sh, I think this flag was changed on this proccess

proto/scripts/rename-conflicting-defs.sh
Thaleszh: This was used for debugging, I echoes some stuff there. Let me remove now

x/feeabstraction/types/query.pb.go
Thaleszh: Because I bumped the cosmos version to v0.53.5

proto/scripts/swagger-protoc.sh
jhelison: Oh, its because the -e is telling the script: `return with stderr if any commands fail`
When its not set, it just executes one command after the other even if one fail

jhelison (approved):

Issue #234, opened 2026-01-15
[BUG] Ignored error parsing validator address & Error swallowed in RemoveExcessFeeds


ngapaxs:
## Bug 1: Ignored error parsing validator address

**Severity:** LOW  
**File:** `x/oracle/abci.go`  
**Line:** 56

### Affected Code

```go
if validator.IsBonded() {
    valPower := validator.GetConsensusPower(powerReduction)
    operator := validator.GetOperator()

    operatorAddr, _ := sdk.ValAddressFromBech32(operator)  // error ignored here

    claim := types.NewClaim(valPower, 0, 0, false, operatorAddr)
    validatorClaimMap[operator] = claim
}
```

### Root Cause

error from `ValAddressFromBech32` ignored with `_`. if parsing fails, operatorAddr ends up nil and claim gets invalid recipient.

### Impact

- claim with nil recipient
- reward distribution could fail
- runs every vote period in EndBlocker

### Recommended Fix

```go
operatorAddr, err := sdk.ValAddressFromBech32(operator)
if err != nil {
    ctx.Logger().Error("failed to parse operator address", "operator", operator, "error", err)
    continue
}
```

### Reference

https://github.com/KiiChain/kiichain/blob/v6.1.0/x/oracle/abci.go#L56

---

## Bug 2: Error swallowed in RemoveExcessFeeds

**Severity:** LOW  
**File:** `x/oracle/keeper/keeper.go`  
**Line:** 272

### Affected Code

```go
for _, denom := range activesToClear {
    err = k.ExchangeRate.Remove(ctx, denom)
    if err != nil {
        return nil  // should be return err
    }
}
```

### Root Cause

typo - returns `nil` instead of `err`. error gets swallowed and caller has no idea something went wrong.

### Impact

- error silently ignored
- stale exchange rates may remain in storage
- called from BeginBlocker

### Recommended Fix

```go
if err != nil {
    return err
}
```

### Reference

https://github.com/KiiChain/kiichain/blob/v6.1.0/x/oracle/keeper/keeper.go#L272

---

## Environment

- Kiichain version: v6.1.0
- Commit: `79bdc96`

## Reporter Declaration

By submitting this issue, I confirm that:

- [x] This report is NOT generated by AI
- [x] I personally tested and verified this issue
- [x] I understand that false or low-effort reports are disqualified from rewards

Thaleszh:
# Bug 1
- Confirmed ✔️ 
[...]

PR #244, opened 2026-01-21
fix(oracle): return error instead of nil in RemoveExcessFeeds

ngapaxs:
# Description

found a typo in `RemoveExcessFeeds` - when `ExchangeRate.Remove` fails it was doing `return nil` instead of `return err`. this swallows the error so the caller never knows removal failed.

split from #234 per maintainer request

## Type of change

# How Has This Been Tested?

ran the existing test to make sure nothing broke:

- [x] TestRemoveExcessFeeds - passes

# PR Checklist:

- [ ] Lint with `make lint-fix`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 - Removed ICS precompile
 
 ## Fixed
+- Return error instead of nil in RemoveExcessFeeds to properly propagate storage errors
 - Fix oracle weighted median threshold to require >50% instead of >=50% for majority
 - Make ERC20 fee conversion rounding explicit by always rounding up to avoid underpayment
 - Handle ValAddressFromBech32 error in oracle EndBlocker [#234](https://github.com/KiiChain/kiichain/issues/234)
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -50,6 +51,7 @@
 -   Bump [EVM](github.com/cosmos/evm) to [v0.4.2](https://github.com/cosmos/evm/releases/tag/v0.4.2)
 
 ## Fixed
+- Return error instead of nil in RemoveExcessFeeds to properly propagate storage errors
```
Thaleszh:
```suggestion
```

Duplicated entry, should just appear on unreleased

x/oracle/keeper/keeper.go
```diff
@@ -269,7 +269,7 @@ func (k Keeper) RemoveExcessFeeds(ctx sdk.Context) error {
 	for _, denom := range activesToClear {
 		err = k.ExchangeRate.Remove(ctx, denom)
 		if err != nil {
-			return nil
+			return err
```
Thaleszh: Ask we talked in the issue, the best would be to just warn the behavior, or investigate deeper the impact of this.

A DB error on remove should be quite rare and hardly widepsread on the chain.

Thaleszh (changes requested):
Need to remove duplicated entry on changelog and review behavior change

CHANGELOG.md
```diff
@@ -52,6 +53,7 @@
...
```
Thaleszh: Still duplicated on changelog

Thaleszh:
Just missing linting

Thaleszh (approved):

PR #271, opened 2026-02-05
Remove unused wasmd precompile

Thaleszh:
# Description

Strips out code from wasmd precompile that is no longer used.

## Type of change

- [x] chore (Updates on dependencies, gitignore, etc)

# How Has This Been Tested?

With pipelines

# PR Checklist:

Make sure each step was done:


app/upgrades/v7_1/constants.go
```diff
@@ -0,0 +1,16 @@
+package v710
+
+import (
+	"github.com/kiichain/kiichain/v7/app/upgrades"
+)
+
+const (
+	// UpgradeName is the name of the upgrade
+	UpgradeName = "v7.1.0"
+)
```
Thaleszh, line 10: So, I don't want to make an upgrade right now, but don't want to forget that we need to remove the precompile if its still there

Thaleszh: Didn't plug the upgrade on the app either

jhelison (approved):

PR #273, opened 2026-02-11
Enforce chain id usage from genesis instead of flag

Thaleszh:
# Description

Always picks up chain ID from genesis instead of flag

## Type of change

# How Has This Been Tested?

Pipelines and locally

# PR Checklist:

Make sure each step was done:

### DIFF
--- a/app/config.go
+++ b/app/config.go
@@ -88,7 +88,7 @@ func ParseChainID(chainID string) (uint64, error) {
 
 	matches := fullChainID.FindStringSubmatch(chainID)
 	if matches == nil || len(matches) != 4 || matches[1] == "" {
-		return 0, fmt.Errorf("chain-id '%s' does not match expected chain ID format: %s_%s-%s", chainID, regexChainID, regexEIP155, regexEpoch)
+		return 0, fmt.Errorf("chain-id '%s' does not match expected chain ID format: <chain name>_<evm chain id>-<epoch separator>, like localhost_1010-1", chainID)
 	}
 
 	// verify that the EIP155 part (matches[2]) is a base 10 integer
--- a/app/config.go
+++ b/app/config.go
@@ -8,8 +8,6 @@ import (
 	"strconv"
 	"strings"
 
-	"github.com/spf13/viper"
-
 	clienthelpers "cosmossdk.io/client/v2/helpers"
 
 	genutiltypes "github.com/cosmos/cosmos-sdk/x/genutil/types"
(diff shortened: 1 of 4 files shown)


cmd/kiichaind/cmd/root.go
```diff
@@ -418,12 +418,15 @@ func (a appCreator) newApp(
 		panic("appOpts is not viper.Viper")
 	}
 
-	existingValue := viperAppOpts.Get(srvflags.EVMChainID)
-	// Check if value exists
-	if existingValue == nil || existingValue == "" {
-		viperAppOpts.Set(srvflags.EVMChainID, kiichain.KiichainID)
+	evmChainIdFlag := cast.ToUint64(viperAppOpts.Get(srvflags.EVMChainID))
+	// Warn if flag is not default
+	if evmChainIdFlag != 262144 {
```
jhelison: Can't we use the value from the evm package? Avoid magic numbers if not. Lets also warn if zero or nil

jhelison: Actually no, scratch the comment above, lets alert if `evmChainIdFlag` is different than `kiichain.KiichainID`

cmd/kiichaind/cmd/root.go
```diff
@@ -418,12 +418,15 @@ func (a appCreator) newApp(
...
+		fmt.Fprintf(os.Stderr, "WARNING: The --%s flag is depecrated. The evm chain ID is being overridden to %d.\n",
```
jhelison: Use the logger from the arguments of the cmd function

app/config.go
```diff
@@ -8,8 +8,6 @@ import (
 	"strconv"
```
jhelison, line 8: Can't comment on the exact lines. But maybe we should enforce the chain ID to follow our format.
This can be done by running the `ParseChainID` on `init` (cmd/kiichaind/cmd/init.go) and on `start` (newApp `cmd/kiichaind/cmd/init.go`). The init here continue to exist, since we have no idea what could leak if we remove it.

cmd/kiichaind/cmd/root.go
Thaleszh: I had a version like that, the issue is the user might still have the evm-id flag and won't be warned.

cmd/kiichaind/cmd/root.go
Thaleszh: Done

jhelison (approved):

PR #279, opened 2026-02-27
feat: re-enable jailing on oracle module

Thaleszh:
# Description
Re-enable jailing on oracle module after fixing the following behaviors
- Denoms with no votes are no longer trigger a miss on everyone
- Overperformers (voting correctly on belowThreshold denoms) no longer get punished
- Validators need to have every above threshold denom correctly to get a success

## Type of change

# How Has This Been Tested?

With added tests and pipelines

# PR Checklist:

Make sure each step was done:

### DIFF
--- a/x/oracle/abci.go
+++ b/x/oracle/abci.go
@@ -157,7 +157,7 @@ func EndBlocker(ctx sdk.Context, k keeper.Keeper) error {
 
 		// Validate miss voting process
 		for _, claim := range validatorClaimMap {
-			if int(claim.WinCount) >= len(voteTargets) {
+			if int(claim.WinCount) == len(voteTargets) {
 				err = k.IncrementSuccessCount(ctx, claim.Recipient)
 				if err != nil {
 					return err
--- a/x/oracle/abci.go
+++ b/x/oracle/abci.go
@@ -141,6 +141,14 @@ func EndBlocker(ctx sdk.Context, k keeper.Keeper) error {
 			Tally(ctx, ballot, params.RewardBand, validatorClaimMap)
 		}
 
+		// Remove any VoteTarget that did not have any vote
+		for denom := range voteTargets {
+			_, inVoteMap := voteMap[denom]
+			if !inVoteMap {
+				delete(voteTargets, denom)
+			}
+		}
+
 		// Validate miss voting process
 		for _, claim := range validatorClaimMap {
 			if int(claim.WinCount) == len(voteTargets) {
--- a/x/oracle/abci.go
+++ b/x/oracle/abci.go
@@ -232,7 +232,7 @@ func BeginBlocker(ctx sdk.Context, k keeper.Keeper) error {
 		return err
 	}
 
-	// Slash who did miss voting over threshold
+	// Slash and jail who did miss voting over threshold
 	// reset miss counter of all validators at the last block of slash window
 	if utils.IsPeriodLastBlock(ctx, params.SlashWindow) {
 		err = k.SlashAndResetCounters(ctx) // slash and jail validator then reset voting counter
--- a/x/oracle/abci.go
+++ b/x/oracle/abci.go
@@ -138,7 +138,12 @@ func EndBlocker(ctx sdk.Context, k keeper.Keeper) error {
 		// Calculate tally for below threshold assets lists
 		for _, denom := range belowThresholdDenoms {
 			ballot := belowThresholdVoteMap[denom]
-			Tally(ctx, ballot, params.RewardBand, validatorClaimMap)
+			// Make a discardable map for belowThreshold votes, so tally changes don't affect scoring
+			discardableClaimMap := make(map[string]types.Claim, len(validatorClaimMap))
+			for k, v := range validatorClaimMap {
+				discardableClaimMap[k] = v
+			}
+			Tally(ctx, ballot, params.RewardBand, discardableClaimMap)
 		}
 
 		// Remove any VoteTarget that did not have any vote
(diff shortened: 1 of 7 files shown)


x/oracle/abci.go
```diff
@@ -138,12 +138,25 @@ func EndBlocker(ctx sdk.Context, k keeper.Keeper) error {
...
 		for _, denom := range belowThresholdDenoms {
 			ballot := belowThresholdVoteMap[denom]
-			Tally(ctx, ballot, params.RewardBand, validatorClaimMap)
+			// Make a discardable map for belowThreshold votes, so tally changes don't affect scoring
+			discardableClaimMap := make(map[string]types.Claim, len(validatorClaimMap))
+			for k, v := range validatorClaimMap {
+				discardableClaimMap[k] = v
+			}
+			Tally(ctx, ballot, params.RewardBand, discardableClaimMap)
+		}
```
Thaleszh: I now don't understand why we tally the denoms under threshold. They are only used to penalize validators, since they are not stored. 

We could remove this part of the code that iterates through them and be done with it.

Thaleszh: I'm failing to see larger consequences of removing them.

To solve the case of validators misvoting on aboveThreshold but getting a successcount on a low threshold (awarding a non miss at the end) I am discarding the usage of the belowThreshold.

We should either keep it as it is or remove it all together

x/oracle/abci.go
```diff
@@ -138,12 +138,25 @@ func EndBlocker(ctx sdk.Context, k keeper.Keeper) error {
...
+			_, inVoteMap := voteMap[denom]
+			if !inVoteMap {
+				delete(voteTargets, denom)
+			}
 		}
 
 		// Validate miss voting process
 		for _, claim := range validatorClaimMap {
-			if int(claim.WinCount) == len(voteTargets) {
+			if int(claim.WinCount) >= len(voteTargets) {
```
Thaleszh: This won't be needed if we remove the below threshold stuff

x/oracle/keeper/slash.go
Thaleszh, line 91: Telemetry with misses, abstain and success

jhelison: Where? This is not telemetry; this is just an event

x/oracle/keeper/slash.go
```diff
@@ -75,6 +75,13 @@ func (k Keeper) SlashAndResetCounters(ctx sdk.Context) error {
 					k.Logger(ctx).Error("failed to slash validator", "operator", operator.String(), "error", err)
 					return true, err
 				}
+
+				// Jail validator
+				err = k.StakingKeeper.Jail(ctx, consAddr)
```
jhelison: Can we only jail if the slashFraction is above zero?

x/oracle/abci.go
jhelison: I don't fully get this tally here, don't we already tally on https://github.com/KiiChain/kiichain/blob/1df0dcba81f6f814d41c9e414f9904071462dd88/x/oracle/abci.go#L111 ?

Maybe we should keep as it is, since on your implementation you discard the data. What would be the impact on keeping as it is?

x/oracle/abci.go
jhelison: Got it, this is where the magic really happens, correct? What really changes here with the `>=`?

x/oracle/abci.go
Thaleszh: The impact would be that we are counting the tally of denoms that have votes but didn't get majority would. We were already discarding their values, but we included those in calculation, punishing for deviation on denoms even without majority.

x/oracle/abci.go
Thaleszh: It makes that voting on below thresholds not being punished (if you get them correctly alongside everything else)

x/oracle/keeper/slash.go
Thaleszh: Oh, I mistook one for the other, let me add it then

x/oracle/keeper/slash.go
Thaleszh: Done

x/oracle/abci.go
Thaleszh: I stripped it out

x/oracle/abci.go
```diff
@@ -128,17 +130,29 @@ func EndBlocker(ctx sdk.Context, k keeper.Keeper) error {
...
+				delete(voteTargets, denom)
+			}
+		}
+
+		// Emit telemetry for all denoms excluded from price aggregation
+		sort.Strings(excludedDenoms)
+		for _, denom := range excludedDenoms {
+			telemetry.SetGaugeWithLabels(
+				[]string{types.ModuleName, "excluded_denom"},
+				1,
```
Thaleszh, line 153: Added telemetry for excluded denoms

x/oracle/abci.go
```diff
@@ -208,6 +222,18 @@ func EndBlocker(ctx sdk.Context, k keeper.Keeper) error {
...
+// abstain, and success counts before the slash window resets them.
+func writeValidatorPenaltyMetrics(ctx sdk.Context, k keeper.Keeper) error {
+	return k.VotePenaltyCounter.Walk(ctx, nil, func(operator sdk.ValAddress, counter types.VotePenaltyCounter) (bool, error) {
+		labels := []metrics.Label{telemetry.NewLabel("validator", operator.String())}
+		telemetry.SetGaugeWithLabels([]string{types.ModuleName, "miss_count"}, float32(counter.MissCount), labels)
+		telemetry.SetGaugeWithLabels([]string{types.ModuleName, "abstain_count"}, float32(counter.AbstainCount), labels)
+		telemetry.SetGaugeWithLabels([]string{types.ModuleName, "success_count"}, float32(counter.SuccessCount), labels)
+		return false, nil
+	})
+}
```
Thaleszh, line 235: Telemetry with abstain, successes and misses of everyone

jhelison (approved):

PR #287, opened 2026-03-17
fix(oracle): add denom length validation and query result limits

giwaov:
## Summary

This PR addresses security vulnerabilities discovered during a security audit of the KiiChain oracle module.

### 1. HIGH: Missing Denom String Length Validation
**Files:** \precompiles/oracle/types.go\, \x/oracle/keeper/query_server.go\

No maximum length validation for denom strings in oracle queries. Extremely long strings could cause memory exhaustion during KVStore operations.

**Fix:** Added \MaxDenomLength\ constant (128 bytes, matching the SDK constraint) with validation in both the precompile parser and the gRPC query handler, ensuring all callers are protected.

### 2. HIGH: Unbounded Memory Allocation in Oracle Queries (DoS)
**File:** \x/oracle/keeper/query_server.go\

Multiple query functions (\ExchangeRates\, \Actives\, \VoteTargets\, \PriceSnapshotHistory\) iterated over all stored data without any limits, allowing attackers to exhaust validator memory.

**Fix:** Added \MaxQueryResults\ (1000) and \MaxSnapshotResults\ (500) constants to limit iteration and prevent memory exhaustion across all list-style queries.

## Impact Assessment

| Severity | Issue | Impact |
|----------|-------|--------|
| HIGH | Missing denom validation | Memory exhaustion via oversized inputs |
| HIGH | Unbounded query iteration | Memory exhaustion, validator DoS |

## Breaking Changes

None. All changes are backward compatible and only affect input validation and resource limits.

PR #288, opened 2026-03-17
fix(oracle): use weight parameter in NewClaim constructor

giwaov:
## Summary

Second round of security audit fixes addressing three code defects found across the oracle and tokenfactory modules.

### 1. Fix NewClaim Weight parameter (x/oracle/types/ballot.go)

The `NewClaim` constructor accepted a `weight` parameter but incorrectly assigned the `power` value to the `Weight` field:

```go
// Before (bug):
Weight: power,

// After (fix):
Weight: weight,
```

This caused the `Weight` field to always equal `Power`, ignoring the intended `weight` argument passed by callers.

### 2. Propagate errors from RemoveExcessFeeds (x/oracle/keeper/keeper.go)

`RemoveExcessFeeds` silently dropped errors from `ExchangeRate.Remove()`, logging them but always returning `nil`. This masked cleanup failures and allowed stale exchange rates to persist without the caller knowing.

The fix tracks removal errors and returns the last error while still attempting all removals:

```go
// Before (bug):
continue  // error silently dropped, function returns nil

// After (fix):
removeErr = err  // error tracked and returned
```

### 3. Nil check in GetAuthorityMetadata (x/tokenfactory/keeper/admins.go)

`GetAuthorityMetadata` called `proto.Unmarshal` with nil bytes when a denom had no authority metadata in the store. `proto.Unmarshal(nil, ...)` silently succeeds and returns an empty struct with `Admin=""`, masking the absence of the denom metadata.

Added a nil check on store bytes to return an explicit error instead of silently returning empty metadata.

## Testing

- Changes are minimal and follow existing code patterns
- Each fix addresses a clear code defect verified by source code review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # CHANGELOG
 
+## Unreleased
+
+### Fixed
+
+- Fix NewClaim constructor assigning power to Weight field instead of the weight parameter (x/oracle/types/ballot.go)
+
 ## v7.1.0-mainnet - 2026-03-13
 
 ### Fixed
--- a/x/oracle/types/ballot.go
+++ b/x/oracle/types/ballot.go
@@ -22,7 +22,7 @@ type Claim struct {
 func NewClaim(power, weight, winCount int64, didVote bool, recipient sdk.ValAddress) Claim {
 	return Claim{
 		Power:     power,
-		Weight:    power,
+		Weight:    weight,
 		WinCount:  winCount,
 		DidVote:   didVote,
 		Recipient: recipient,


Thaleszh:
Hello @giwaov, we reviewed the issues you opened and some of them are not real, but some are!

Can you take a look again on the issues that remained open? Also, feel free to discuss any of the issues we closed.

After that, you will be needing to refactor both this PR and your other one, since they include changes we don't agree with (solving non-issues).

giwaov:
Hey @Thaleszh, thanks for taking the time to review everything. Appreciate the thorough feedback.

I've gone through the closed issues and the reasoning makes sense  the proto types guaranteeing non-nil values in the store (#294) and the intentional best-effort design for RemoveExcessFeeds (#295) are things I missed during the audit. Good to know.

I've updated both PRs accordingly:

- **This PR (#288)** now only contains the NewClaim Weight fix (ballot.go)
- **PR #287** now only contains the denom length validation (types.go) and the unbounded query limits (query_server.go)

The rejected changes have been removed from both branches. Let me know if anything else needs adjusting. Thank you.

Thaleszh:
Can you rename the PR to reflect the changes? 
Also add to changelog the fix done on the PR, under unreleased.

giwaov:
Hey @Thaleszh, done!

- **PR title** updated to reflect the current scope
- **Changelog** entry added under Unreleased for the NewClaim weight fix

Let me know if anything else needs adjusting, thank you.

Thaleszh (approved):

PR #303, opened 2026-03-31
fix: ensure nativeOracleDenom is whitelisted and exists in voteTargets

Thaleszh:
# Description

Fixes:
- Ensure nativeOracleDenom is whitelisted and exists in voteTargets

## Type of change

# How Has This Been Tested?

Added regression tests

# PR Checklist:

Make sure each step was done:

### DIFF
--- a/x/feeabstraction/keeper/msg_server.go
+++ b/x/feeabstraction/keeper/msg_server.go
@@ -50,11 +50,14 @@ func (ms MsgServer) UpdateParams(ctx context.Context, msg *types.MsgUpdateParams
 	if err != nil {
 		return nil, sdkerrors.ErrInvalidRequest.Wrapf("failed to get oracle vote targets: %s", err)
 	}
-	voteTargetSet := make(map[string]struct{}, len(voteTargets))
-	for _, denom := range voteTargets {
-		voteTargetSet[denom] = struct{}{}
+	found := false
+    for _, denom := range voteTargets {
+		if denom == msg.Params.NativeOracleDenom {
+            found = true
+            break
+        }
 	}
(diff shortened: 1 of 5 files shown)


x/feeabstraction/keeper/msg_server.go
```diff
@@ -40,10 +40,33 @@ func (ms MsgServer) UpdateParams(ctx context.Context, msg *types.MsgUpdateParams
...
+	if err != nil {
+		return nil, sdkerrors.ErrInvalidRequest.Wrapf("failed to get oracle vote targets: %s", err)
+	}
+	voteTargetSet := make(map[string]struct{}, len(voteTargets))
+	for _, denom := range voteTargets {
+		voteTargetSet[denom] = struct{}{}
+	}
+	if _, ok := voteTargetSet[msg.Params.NativeOracleDenom]; !ok {
+		return nil, sdkerrors.ErrInvalidRequest.Wrapf("native oracle denom %s is not registered as an oracle vote target", msg.Params.NativeOracleDenom)
+	}
```
jhelison: Why iterate then do the look?
```suggestion
	found := false
    for _, denom := range voteTargets {
		if denom == msg.Params.NativeOracleDenom {
            found = true
            break
        }
	}
	if !found {
		return nil, sdkerrors.ErrInvalidRequest.Wrapf("native oracle denom %s is not registered as an oracle vote target", msg.Params.NativeOracleDenom)
	}
```

Thaleszh: I was doing something else with the map, but no longer needed

jhelison (approved):

jhelison (approved):

PR #304, opened 2026-03-31
fix: apply whitelist denoms on feeabstration genesis

Thaleszh:
# Description

Fixes:
- Apply whitelist denoms on feeabstration genesis 

## Type of change

# How Has This Been Tested?

Added tests for each case

# PR Checklist:

Make sure each step was done:


x/feeabstraction/keeper/msg_server.go
```diff
@@ -106,8 +107,17 @@ func (ms MsgServer) UpdateFeeTokens(ctx context.Context, msg *types.MsgUpdateFee
 		}
 	}
 
+	// Zero out all token prices so BeginBlocker adopts the current TWAP immediately
+	// rather than being clamped from a stale governance-submitted price.
+	resetItems := make([]types.FeeTokenMetadata, len(msg.FeeTokens.Items))
```
jhelison: Should not the un-update field be remove the proto too?

Thaleszh: Yeah, that's what I mentioned about it needing a bit more restructure. This would mean changing a lot of tests

jhelison (approved):

PR #305, opened 2026-04-01
fix: ensure reward community pool has funds

Thaleszh:
# Description

Ensures that the community pool has enough funds before taking from it.

## Type of change

# How Has This Been Tested?

Has a regression test

# PR Checklist:

Make sure each step was done:

### DIFF
--- a/x/rewards/keeper/abci.go
+++ b/x/rewards/keeper/abci.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 
 	"cosmossdk.io/math"
+
 	"github.com/cosmos/cosmos-sdk/telemetry"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 
--- a/x/rewards/keeper/abci.go
+++ b/x/rewards/keeper/abci.go
@@ -42,7 +42,7 @@ func (k Keeper) BeginBlocker(ctx sdk.Context) error {
 	// If active and there is no previous time stamp, set it as current block's and skip this time
 	if schedule.LastReleaseTime.IsZero() {
 		schedule.LastReleaseTime = ctx.BlockTime()
-		return k.haltSchedule(ctx, schedule, err)
+		return k.ReleaseSchedule.Set(ctx, schedule)
 	}
 
 	// Calculate the amount to distribute this block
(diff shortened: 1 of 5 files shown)


x/rewards/keeper/genesis.go
```diff
@@ -19,6 +20,21 @@ func (k Keeper) InitGenesis(ctx sdk.Context, data types.GenesisState) {
...
+	bankBalance := k.bankKeeper.GetBalance(ctx, moduleAddr, denom)
+	poolAmount := data.RewardPool.CommunityPool.AmountOf(denom).TruncateInt()
+	if !bankBalance.Amount.Equal(poolAmount) {
+		k.Logger(ctx).Error(
+			"rewards module bank balance does not match CommunityPool accounting at genesis",
+			"denom", denom,
+			"bank_balance", bankBalance.Amount,
+			"community_pool", poolAmount,
+		)
+	}
```
jhelison, line 37: This was requested on the audit?

x/rewards/keeper/abci.go
```diff
@@ -50,13 +54,24 @@ func (k Keeper) BeginBlocker(ctx sdk.Context) error {
 	// Set up coins
 	coinsToDistribute := sdk.NewCoins(amountToDistribute)
 
+	// Verify the CommunityPool has sufficient balance for the denom before transferring
+	poolAmount := rewardPool.CommunityPool.AmountOf(amountToDistribute.Denom)
+	if math.LegacyNewDecFromInt(amountToDistribute.Amount).GT(poolAmount) {
+		return fmt.Errorf("community pool has insufficient balance for %s: pool has %s, need %s",
+			amountToDistribute.Denom, poolAmount, amountToDistribute.Amount)
+	}
```
jhelison, line 70: Can this possibly cause problems and halt the chain?

x/rewards/keeper/genesis.go
Thaleszh: Yes:

During genesis validation or initialization, optionally verify that the rewards module account balance for the configured token denom matches the corresponding CommunityPool balance.

x/rewards/keeper/abci.go
Thaleszh: Previously it would panic in this scenario, now we're returning an error.

The same happens on the negative check after safe sub.

I will switch both to error logs and turn off the schedule if this happens

jhelison (dismissed):

jhelison (approved):

PR #322, opened 2026-04-08
fix: ensure price update follows oracle module after UpdateTokenMetadata in fee abstraction

Thaleszh:
# Description

- Ensure price update follows oracle module after UpdateTokenMetadata in fee abstraction
- Remove price input from updateTokenMetadata, always using next TWAP cycle

## Type of change

# How Has This Been Tested?

Updated tests

# PR Checklist:

Make sure each step was done:


x/feeabstraction/keeper/msg_server.go
```diff
@@ -103,14 +104,22 @@ func (ms MsgServer) UpdateFeeTokens(ctx context.Context, msg *types.MsgUpdateFee
...
 		voteTargetMap[denom] = struct{}{}
 	}
-	for _, feeToken := range msg.FeeTokens.Items {
+	for _, feeToken := range msg.Tokens.Items {
 		if _, ok := voteTargetMap[feeToken.OracleDenom]; !ok {
 			return nil, sdkerrors.ErrInvalidRequest.Wrapf("fee token denom %s is not registered on the oracle module", feeToken.OracleDenom)
 		}
 	}
 
+	// Zero out all token prices so BeginBlocker adopts the current TWAP immediately
```
jhelison, line 113: Where is the logic that zero out prices should take the oracle prices already?

Thaleszh: It happens here, when previous price is at zero (the one currently stored on the module)

https://github.com/KiiChain/kiichain/blob/fix/remove-price-update-feetoken/x/feeabstraction/types/fee.go#L55

jhelison: Can we add to the docstring pointing to this code?

Thaleszh: done

jhelison (approved):

PR #342, opened 2026-06-19
fix: add authz guarded router

mattkii:
# Description

This PR closes an EVM ante-bypass vector reachable through the authz module. Cosmos SDK's authz keeper unwraps `cosmos.authz.v1beta1.MsgExec` and dispatches the nested messages directly through the app message router, skipping kiichain's Cosmos ante handler. Because the EVM-specific protections (`RejectMessagesDecorator` and `AuthzLimiterDecorator`) only run in the ante path, a nested `cosmos.evm.vm.v1.MsgEthereumTx` could be routed into the EVM `MsgServer` outside of the EVM ante/nonce/fee/block-gas checks. This is exploitable on any execution path that dispatches authz without running the ante (e.g. gov proposal execution, CosmWasm contract dispatch, and the ICA host if it is ever re-enabled), and the no-grant self-grantee variant requires no authz grant since `MsgEthereumTx.From` can be spoofed independently of the recovered Ethereum signer.

The fix wraps the `baseapp.MessageRouter` passed into the `AuthzKeeper` with a guard that refuses to return a handler for `MsgEthereumTx` (and refuses it via `HandlerByTypeURL`, which also blocks creating authz grants for that type). The guard is scoped to the AuthzKeeper reference only, so baseapp tx routing and all other modules keep using the unwrapped router. Because nested `MsgExec`s re-enter the same guarded router at every level, this catches `MsgEthereumTx` at any nesting depth. There is no functional regression: executing `MsgEthereumTx` through authz was already rejected at the ante for normal transactions, so no legitimate path relied on it. The change is contained to kiichain; no changes to the EVM module are required.

Dependencies: none.

## Type of change

# How Has This Been Tested?

- [x] `go build ./app/...` passes on the repo's Go 1.24 toolchain.
- [x] Authz `MsgExec` wrapping `MsgEthereumTx` is rejected with `ErrUnauthorized` at dispatch (direct and nested), while non-EVM authz `MsgExec` continues to execute normally.

# PR Checklist:

### DIFF
--- a/app/keepers/authz_router.go
+++ b/app/keepers/authz_router.go
@@ -1,0 +1,52 @@
+package keepers
+
+import (
+	errorsmod "cosmossdk.io/errors"
+
+	"github.com/cosmos/cosmos-sdk/baseapp"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	errortypes "github.com/cosmos/cosmos-sdk/types/errors"
+
+	evmtypes "github.com/cosmos/evm/x/vm/types"
+)
+
+// evmMsgTypeURL is the type URL of MsgEthereumTx, cached once so it can be
+// compared cheaply when routing messages.
+var evmMsgTypeURL = sdk.MsgTypeURL(&evmtypes.MsgEthereumTx{})
+
+// evmRejectingMessageRouter wraps a baseapp.MessageRouter and prevents EVM
+// transactions (MsgEthereumTx) from being executed through authz. EVM messages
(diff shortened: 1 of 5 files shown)


app/keepers/authz_router.go
jhelison, line 1: Can you docstring this file?

mattkii: done

jhelison (dismissed):

jhelison (dismissed):

jhelison (approved):

PR #345, opened 2026-06-25
fix: cosmwasm evm query path repeatable undercharged evm exec

mattkii:
# Description

Fixes a gas-accounting vulnerability in the CosmWasm→EVM query bridge
(`wasmbinding/evm/queries.go`). Any CosmWasm contract could call the ERC20
information/balance/allowance queries against an arbitrary EVM contract, and each
sub-call was issued with `gasCap=nil`. The underlying EVM helper ignores a nil cap
and falls back to `DefaultGasCap` (25M), so an attacker-controlled ERC20 could force
large, repeatable internal EVM execution while the SDK gas meter was charged only the
post-refund amount — enabling sustained, undercharged validator compute.

This PR makes every internal ERC20 query sub-call pass a real gas cap bounded by the
transaction's remaining SDK gas (and `DefaultGasCap`), mirroring the existing
`buildEthCallRequest` behaviour used by the generic `EthCall` path.

Changes:
- Add `queryGasCap(ctx)` helper returning `min(ctx.GasMeter().GasRemaining(), DefaultGasCap)`.
- Pass `queryGasCap(ctx)` instead of `nil` to all six sub-calls: `decimals`, `name`,
  `symbol`, `totalSupply`, `balanceOf`, `allowance`.

Dependencies: full mitigation also relies on the companion `KiiChain/evm` PR
(`fix/wasm-repeatable-undercharged-internal-evm-exec`) where `CallEVMWithData` now
honours the gas cap and bills the SDK meter for actual pre-refund EVM work. That fork
version must be published and the `replace github.com/cosmos/evm => github.com/KiiChain/evm`
directive in `go.mod` bumped for the fix to take full effect on-chain.

## Type of change

# How Has This Been Tested?

- [x] `go test -tags=test ./wasmbinding/evm/...` — all query binding tests pass (incl. ERC20 information/balance/allowance and EthCall gas tests).
- [x] `go build ./wasmbinding/...` — compiles cleanly.

# PR Checklist:

- [x] Lint with `make lint-fix` (golangci-lint v2.7.2, 0 issues)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -47,6 +47,10 @@ jobs:
         if: env.GIT_DIFF
         run: |
           go test -v -coverprofile=profile.txt -covermode=atomic -coverpkg=./... $(go list ./... | grep -v -e '/tests/e2e' | grep -v -e '/tests/interchain') | grep -v "store.go:" | grep -v "mutable_tree.go:"
+      - name: wasmbinding coverage (test build tag)
+        if: env.GIT_DIFF
+        run: |
+          go test -tags=test -coverprofile=profile-wasmbinding.txt -covermode=atomic ./wasmbinding/...
       - uses: actions/upload-artifact@v4
         if: env.GIT_DIFF
         with:
@@ -56,7 +60,7 @@ jobs:
         uses: codecov/codecov-action@v4
         if: env.GIT_DIFF
         with:
-          files: ./profile.txt
+          files: ./profile.txt,./profile-wasmbinding.txt
           token: ${{ secrets.CODECOV_TOKEN }}
 
   repo-analysis:
--- a/wasmbinding/evm/queries.go
+++ b/wasmbinding/evm/queries.go
@@ -8,6 +8,7 @@ import (
 	"math/big"
 
 	wasmvmtypes "github.com/CosmWasm/wasmvm/v3/types"
+	"github.com/ethereum/go-ethereum/accounts/abi"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/core/vm"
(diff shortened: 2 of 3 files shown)


.github/workflows/test.yml
jhelison, line 1: Why were these changes required?

wasmbinding/evm/queries.go
```diff
@@ -303,6 +303,14 @@ func handleRevertError(vmError string, ret []byte) error {
 	return nil
 }
 
+// queryGasCap returns the gas cap to apply to internal EVM query sub-calls.
+// It is bounded by the transaction's remaining SDK gas so that a CosmWasm
+// contract cannot force more EVM computation than its transaction pays for,
+// and additionally capped by DefaultGasCap to match the eth_call behaviour.
+func queryGasCap(ctx sdk.Context) *big.Int {
+	return new(big.Int).SetUint64(min(ctx.GasMeter().GasRemaining(), emvconfig.DefaultGasCap))
+}
```
jhelison: Interesting approach. Would it not be better to just keep the GasRemaining? I'm saying this because the evm side with the new changes, we kind of already take the minimum

wasmbinding/evm/queries.go
jhelison, line 1: Check if the suggestion at https://github.com/KiiChain/evm/pull/18#discussion_r3481296142 make sense

.github/workflows/test.yml
mattkii: the wasmbinding tests need -tags=test.  The main coverage run doesn't pass -tags=test, Codecov flagged this PR's queries.go lines as 0% patch. The extra step runs only ./wasmbinding/... with the tag and uploads a second profile, without changing how any other package runs.

wasmbinding/evm/queries.go
mattkii: makes sense, good call.

jhelison (approved):

PR #346, opened 2026-06-30
fix: prevent rewards BeginBlocker chain halt on bank transfer failure

mattkii:
# Description

Fixes a chain-liveness bug in the `x/rewards` module. In `BeginBlocker`, a failed `SendCoinsFromModuleToModule` (rewards module to fee collector) returned the error directly, which aborts `FinalizeBlock` and halts every validator. The other failure paths in the same function (`CalculateReward` error, insufficient community pool, negative `SafeSub`) already deactivate the schedule gracefully via `haltSchedule`; the bank transfer was the one path that didn't. The trigger is a divergence between the rewards module's internal `CommunityPool` accounting and the module account's actual bank balance (e.g. misconfigured genesis or prior state inconsistency).

Changes:

- Route the bank transfer failure through `haltSchedule`, so a transfer error deactivates the release schedule and the block still commits (matches the other error paths). The transfer happens before any `RewardPool`/schedule state writes, so no partial accounting is persisted on failure.
- Add `Keeper.ValidateModuleAccounting` (module bank balance must cover the `CommunityPool`) as a single source of truth, and run it at **genesis** (logged warning, module InitGenesis ordering can fund the account after this runs, so a hard failure would false positive) and during the **v7.2 upgrade** handler (logged after migrations) to surface an accounting/bank divergence early.

## Type of change

# How Has This Been Tested?

- [x] `go test -tags=test -run 'TestKeeperTestSuite/TestEndBlocker' ./x/rewards/keeper/...` — added a `"bank transfer failure - halts schedule gracefully"` case that sets `CommunityPool` accounting above the module account's actual balance, forcing the release transfer to fail; asserts `BeginBlocker` returns no error and the schedule becomes inactive (this case fails without the fix).
- [x] `go test -tags=test -run 'TestKeeperTestSuite/TestValidateModuleAccounting' ./x/rewards/keeper/...` — covers empty pool (ok), funded pool matching balance (ok), and accounting inflated beyond balance (error).
- [x] `go build ./x/rewards/... ./app/...` — compiles cleanly.

# PR Checklist:

- [x] Lint with `make lint-fix` (golangci-lint v2.7.2, 0 issues on `./x/rewards/...` and `./app/upgrades/...`)

### DIFF
--- a/app/upgrades/v7_2/upgrade.go
+++ b/app/upgrades/v7_2/upgrade.go
@@ -17,7 +17,7 @@ import (
 func CreateUpgradeHandler(
 	mm *module.Manager,
 	configurator module.Configurator,
-	_ *keepers.AppKeepers,
+	appKeepers *keepers.AppKeepers,
 ) upgradetypes.UpgradeHandler {
 	return func(c context.Context, _ upgradetypes.Plan, vm module.VersionMap) (module.VersionMap, error) {
 		ctx := sdk.UnwrapSDKContext(c)
@@ -27,6 +27,10 @@ func CreateUpgradeHandler(
 		vm, err := mm.RunMigrations(ctx, configurator, vm)
 		if err != nil {
 			return vm, err
+		}
+
+		if err := appKeepers.RewardsKeeper.ValidateModuleAccounting(ctx); err != nil {
+			ctx.Logger().Error("rewards module accounting invariant violated during upgrade", "error", err)
 		}
 
 		ctx.Logger().Info("Upgrade v7.2.0 complete")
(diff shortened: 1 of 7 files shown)


app/upgrades/v7_2/upgrade.go
```diff
@@ -17,7 +17,7 @@ import (
 func CreateUpgradeHandler(
```
jhelison, line 17: Why the changes to an already existing upgrade?

mattkii: Good catch.  I was mainly testing it for upgrade for the invariant.  I've removed it.

jhelison (approved):

PR #353, opened 2026-07-06
fix: Expedited Governance Proposal Whitelist Bypass via authz Wrapping

mattkii:
# Description

The `GovExpeditedProposalsDecorator` only inspected top level messages, so a non-whitelisted `MsgSubmitProposal` marked as expedited could slip through by being wrapped inside an `authz.MsgExec`. That let non-whitelisted proposal types use the shortened expedited voting window that governance policy intended to reserve for the standard timeline.

The fix unwraps `authz.MsgExec` recursively and applies the same expedited whitelist validation to the wrapped proposals, matching the pattern already used by `GovVoteDecorator`.

No new dependencies.

## Type of change

# How Has This Been Tested?

```bash
go test -tags=test ./ante/ -run TestGovExpeditedProposalsDecorator -v
```

- [x] A non-whitelisted expedited proposal wrapped in `authz.MsgExec` (and nested `exec(exec(...))`) is now rejected
- [x] A whitelisted expedited proposal in authz still passes, and non-expedited proposals in authz are unaffected

# PR Checklist:


ante/gov_expedited_ante.go
jhelison, line 1: Can you document the new functions? Avoid the paragraphs that AI usually writes

jhelison (approved):

jhelison (approved):

PR #367, opened 2026-07-31
feat: rewards utility distribution

mattkii:
# Description

Replaces the timed linear `ReleaseSchedule` in `x/rewards` with continuous inflation-based emissions driven by bonded ratio.

Rewards still come from the prefunded pool → `fee_collector` → `x/distribution` path (begin-block order unchanged). Per-block amount is now:

`clamp((1 − bonded/goal) × 0.13 × bonded, min, max) × supply_base × Δt / year`

capped at remaining pool balance (emits until the pool runs dry).

**Removed:** `ReleaseSchedule`, `MsgChangeSchedule`, end-date / total-budget schedule semantics, and the release-schedule query/CLI.

**Added:** staking `BondedRatio` dependency; gov params `goal_bonded`, `inflation_min`, `inflation_max`, `supply_base` (default `0` = off); hardcoded `inflation_rate_change = 0.13`; `last_release_time` / `total_released` on `RewardPool`; richer `reward_distributed` events (`inflation_rate`, `bonded_ratio`).

Enable path: fund pool (`MsgFundPool`) → gov `MsgUpdateParams` with `supply_base > 0`. Disable: set `supply_base` to `0`.

Consensus version bumped to `2`. No new external dependencies.

## Type of change

> Note: this is a **breaking** module state/API change for `x/rewards` (schedule removed; params/state shape changed). Treat as a feature for the utility-rewards design, not a drop-in non-breaking add.

# How Has This Been Tested?

- [x] Unit: `go test -tags=test ./x/rewards/...` (curve math, BeginBlocker, msgs, queries, params, genesis)
- [x] Targeted: `go test -tags=test ./x/rewards/types -run 'TestCalculate'` and `./x/rewards/keeper -run 'TestKeeperTestSuite/TestBeginBlocker'`
- [x] E2E: `SKIP_IBC_TESTS=true go test -tags=test ./tests/e2e -run 'TestIntegrationTestSuite/TestRewards'` (fund pool → gov set `supply_base` → pool decreases / validator rewards increase)
- [x] Manual local: `./contrib/scripts/test_rewards_manual.sh` (fund → gov set `supply_base` → pool decreases / `total_released` increases)

# PR Checklist:

Make sure each step was done:


.github/workflows/test.yml
```diff
@@ -51,16 +51,23 @@ jobs:
         if: env.GIT_DIFF
         run: |
           go test -tags=test -coverprofile=profile-wasmbinding.txt -covermode=atomic ./wasmbinding/...
+      - name: rewards coverage (test build tag)
+        if: env.GIT_DIFF
+        run: |
+          go test -tags=test -coverprofile=profile-rewards.txt -covermode=atomic ./x/rewards/...
```
jhelison, line 57: Rewards as not being covered by the topic above? Why the new entry here?

contrib/scripts/test_rewards_manual.sh
```diff
@@ -0,0 +1,297 @@
+#!/usr/bin/env bash
```
jhelison, line 1: Can't this be turned into 2e2 test?

proto/kiichain/rewards/v1beta1/types.proto
```diff
@@ -8,41 +8,24 @@ import "amino/amino.proto";
...
   ];
-}
\ No newline at end of file
+
+  // Timestamp of the last emission release
+  google.protobuf.Timestamp last_release_time = 2 [
+    (gogoproto.stdtime) = true,
+    (gogoproto.nullable) = false,
+    (gogoproto.moretags) = "yaml:\"last_release_time\""
+  ];
```
jhelison, line 24: Isn't release per block? This would mean that we don't really need the last release since its per block

proto/kiichain/rewards/v1beta1/params.proto
```diff
@@ -2,12 +2,35 @@ syntax = "proto3";
...
+  string inflation_max = 4 [
+    (gogoproto.customtype) = "cosmossdk.io/math.LegacyDec",
+    (gogoproto.nullable) = false
+  ];
+
+  // Base used to size annual provisions (in token base units)
+  string supply_base = 5 [
+    (gogoproto.customtype) = "cosmossdk.io/math.Int",
+    (gogoproto.nullable) = false
+  ];
```
jhelison, line 35: What is the supply base?

CHANGELOG.md
```diff
@@ -9,7 +9,8 @@
 
 ### Added
 
-- Emit `update_params`, `fund_pool`, `change_schedule`, and `reward_distributed` events from x/rewards
+- Replace timed linear `ReleaseSchedule` emissions in `x/rewards` with continuous inflation-based utility rewards driven by bonded ratio: `inflation = clamp((1 - bonded/goal) × 0.13 × bonded, min, max)`, amount = `inflation × supply_base × Δt / year`, capped at remaining pool balance (emits until pool runs dry). Adds staking `BondedRatio` dependency and gov params `goal_bonded`, `inflation_min`, `inflation_max`, `supply_base` (default `0` disables emissions); `inflation_rate_change` is hardcoded at `0.13`. Enable via `MsgFundPool` + gov `MsgUpdateParams`
```
jhelison, line 12: These are being added to the unreleased section. We need to updade the changelog with the correct released versions

x/rewards/README.md
```diff
@@ -1,149 +1,60 @@
...
-- **Changing denoms**: The token denom being distributed can be changed, but it was not meant to support multiple distributions at the same time. It was merely set as a future possibility for other tokens to be rewarded instead of the native token. If the denom needs to be changed, the correct approach is to wait for the current schedule to end so no tokens are left over, then change the base denom and create a new schedule. Otherwise, some tokens may be left over in the pool.
-  - If a denom is changed, the `QueryParamsRequest` will return this new denom, even if there is still a current schedule on the old denom.
-- **Leftover tokens**: If tokens are left over by any schedule due to overrides, they can be distributed in the future with a new schedule, but not recovered. With this in mind, the recommended approach when changing a schedule is to first halt it, then include its leftover funds in the next schedule, assuming it's on the same token.
-
-## Messages
+```
+inflation = clamp((1 - bondedRatio/goalBonded) × 0.13 × bondedRatio, inflationMin, inflationMax)
+amount    = inflation × supplyBase × elapsedNs / nsPerYear
+pay       = min(amount, poolBalance)
```
jhelison, line 19: Isn't the 0.13 a param? Also what does `bondedRatio, inflationMin, inflationMax` means? Should not this be a single param?
What is the supply base?

x/rewards/README.md
```diff
@@ -1,149 +1,60 @@
...
+```
 
-### FundPool
+`inflation_rate_change` is a fixed constant (`0.13`). `supply_base` defaults to `0` (emissions off).
```
jhelison, line 22: Why inflation rate change is a fixed constant?

x/rewards/module.go
```diff
@@ -40,7 +40,7 @@ var (
 )
 
 // ConsensusVersion defines the current x/rewards module consensus version.
-const ConsensusVersion = 1
+const ConsensusVersion = 2
```
jhelison, line 43: Besides the consensus version, should not we have the upgrade set on the module?
Here you can see one example:
- https://github.com/KiiChain/kiichain/blob/7689cd3141874cdd3c754f1ad31debf8842a89f2/x/tokenfactory/migrations/v2/migrate.go

x/rewards/types/reward.go
```diff
@@ -8,36 +8,64 @@ import (
...
-	totalDurationStamp := schedule.EndTime.Sub(schedule.LastReleaseTime) // Remaining release period
+	if p.SupplyBase.IsZero() {
+		return zero, math.LegacyZeroDec()
+	}
 
-	// Use nanoseconds for precise duration calculations
-	timeElapsedNs := math.NewInt(timeElapsedStamp.Nanoseconds()).BigInt()
-	totalDurationNsBig := math.NewInt(totalDurationStamp.Nanoseconds()).BigInt()
+	inflation := CalculateInflation(bondedRatio, p)
+	annualProvision := inflation.MulInt(p.SupplyBase)
```
jhelison, line 57: How the supplyBase work here? It seems that the naming is misleading

x/rewards/types/reward.go
```diff
@@ -8,36 +8,64 @@ import (
...
-	if remaining.IsZero() {
-		return remaining, nil
-	}
+// SecondsPerYear is the reference year length (365 days) used to convert an
+// annual emission rate into a per-block provision.
+const SecondsPerYear = 31_536_000
+
+// inflationRateChange is the max annual swing of the emission rate.
+// Hardcoded (not a gov param) so emission math stays consensus-stable.
+var inflationRateChange = math.LegacyNewDecWithPrec(13, 2) // 0.13
```
jhelison, line 17: This is one of the params that we use to control the inflation, it should be a on-chain param instead of a hardcoded value

x/rewards/types/reward_test.go
```diff
@@ -8,167 +8,111 @@ import (
...
-// TestCalculateReward verifies the CalculateReward function across a comprehensive
-// set of scenarios: fully-exhausted schedules, linear proportional releases at
-// various elapsed fractions, partial prior releases, nanosecond-precision
-// sub-second durations (regression for issue #267), and the edge case where
-// EndTime equals LastReleaseTime, which must release the full remaining amount
-// immediately rather than returning an error.
-func TestCalculateReward(t *testing.T) {
-	now := time.Now()
-	denom := "akii"
+func TestCalculateInflation(t *testing.T) {
```
jhelison, line 14: This test replicate the same numbers that we used on our spreadsheet?

x/rewards/keeper/abci.go
```diff
@@ -11,122 +9,118 @@ import (
...
-// the error never reaches FinalizeBlock and halts the chain
-func (k Keeper) haltSchedule(ctx sdk.Context, schedule types.ReleaseSchedule, err error) error {
-	k.Logger(ctx).Error("halting release schedule due to error", "error", err)
-	schedule.Active = false
-	return k.ReleaseSchedule.Set(ctx, schedule)
-}
-
-// BeginBlocker calculates reward amt and sends it to the distribution pool
+// BeginBlocker calculates the inflation-based reward amount and sends it to the fee collector.
 func (k Keeper) BeginBlocker(ctx sdk.Context) error {
```
jhelison, line 13: Did not fully get the advanceLastReleaseTime. I'm asking because I see like this:
- On every block we should release the amount based on the on-chain params, so the tracking using the last released feels weird here, can you explain it better to me?

.github/workflows/test.yml
mattkii: Same reason as wasmbinding: rewards keeper tests are behind -tags=test. The main coverage job runs without that tag, so those packages don’t show up in Codecov unless we run a tagged profile (and upload it). Happy to fold into the primary step with -tags=test instead if you prefer one job.

contrib/scripts/test_rewards_manual.sh
mattkii: Yep, it is already.  `tests/e2e/e2e_rewards_test.go` (`TestRewards` then `fund pool` then `gov supply_base` then `assert pool drain / rewards`). `contrib/scripts/test_rewards_manual.sh` is optional local smoke without the full e2e harness. Fine to drop the script if you’d rather keep only CI e2e.

proto/kiichain/rewards/v1beta1/types.proto
mattkii: So i've set it so emission is wall-clock proportional: amount ∝ Δt / year, not “1/N of a block.” Block intervals i figured is around ~1-3s and we also skip intervals (empty pool, bank/staking errors, etc.). Without a timestamp, those skips would either under pay forever or dump a huge catch up later. We stamp time on first active block, then each successful (or intentionally skipped) interval.

CHANGELOG.md
mattkii: Since this isn't released yet, i figured it belongs under Unreleased until we cut the release (e.g. v7.4.0). I’ll rebase onto latest main so already-released bullets aren’t duplicated under Unreleased. If we want a dedicated ## vX.Y.Z section now, let me know and i can change the target version.

x/rewards/module.go
mattkii: Agree. `x/rewards/migrations/v2` deletes legacy ReleaseSchedule, backfills default inflation params when upgrading from denom-only v1 params, and is registered via `cfg.RegisterMigration(..., 1, ...)`. Tests included.

x/rewards/types/reward.go
mattkii: Fair. It’s an emission scale knob, not` bank.GetSupply()`. Docs clarified; happy to rename the proto field.  Maybe `emission_base`?  `provision_base`?

x/rewards/types/reward.go
mattkii: Originally hardcoded so the curve shape couldn’t drift without a binary upgrade. Happy to move it to gov params (default 0.13).   I’ll do that in a follow-up commit (proto + validation + migration default).

x/rewards/keeper/abci.go
mattkii: On a normal block we release `f(params, bonded) × Δt` and `set last_release_time = now.` If we skip (staking read fail, bank send fail, empty pool) but leave the old timestamp, the next success uses a huge `Δt` and can dump up to the whole pool in one block. Advancing the clock on those skips means “this interval paid 0; don’t accrue it.” Successful pays still update time after the transfer.

proto/kiichain/rewards/v1beta1/params.proto
mattkii: Not chain total supply. It’s a gov notional base that sizes annual provisions: `annual_provision = inflation × supply_base`, then prorated by `Δt`. Zero disables emissions (feature flag). Prefunded pool is still the actual source of funds. `supply_base` only sets the drip rate. Clarified in proto/params comments.

x/rewards/README.md
mattkii: Here are the the params:
- `bondedRatio` - fraction of tokens currently staked (from staking)
- `goalBonded` - target stake ratio (gov param, default 0.67)
- `inflationMin` / `inflationMax` - floor/ceiling on the emission rate (gov params)
- `0.13` - `inflation_rate_change` (curve steepness). Hardcoded today; moving it to a gov param per your other comment
- `supply_base` - gov knob for drip speed from the prefunded pool (`yearly ≈ inflation × supply_base`), not chain total supply. `0` = off

I’ve update the README to explain this out more clearly.

x/rewards/README.md
mattkii: Originally locked in the binary so the curve couldn’t change without an upgrade. Agree it should be gov-controlled.  I’ll add it as an on-chain param (default `0.13`).

x/rewards/types/reward_test.go
mattkii: Yep. Added vectors from the sheet in  (`goalBonded=0.67`, `inflationRateChange=0.13`, `supplyBase=1.8e9`). They assert the inflation column and the truncated annual provision values across the bonded ratio table.

x/rewards/types/reward.go
jhelison: Apply it

x/rewards/keeper/abci.go
jhelison: Uhm, but there is no deltaT on Cosmos Mint module implementation. All that we have are the releases based on the math formula. If we say that the release is 10%, then it would be `(Supply * 10%) / Total blocks in a year`. This happens without having any deltaT.
It seems that this is related to my misunderstanding of the behavior of supplyBase

CHANGELOG.md
jhelison: Yes, also referring to the recent releases, we did not move the Unreleased to the correct versions on the changelog

.github/workflows/test.yml
jhelison: But why are we using the test tag on the testing files instead of scoping them as test files?

mattkii: They are scoped as test files, every one of them is `_test.go`. The tag isn't gating our test code, it's required by the cosmos/evm dependency.  Any test that boots the full app (our apptesting/helpers wire the EVM module) calls `EVMConfigurator.ResetTestConfig()`. In cosmos/evm that function is a stub that panics with "this is only implemented with the 'test' build flag",  The working implementation only exists in their `test`-tagged file. So without `-tags=test` the rewards keeper suite panics at setup rather than failing to compile.
That's why `make test` and `make test-unit` already pass `-tags=test`. The CI coverage step is the one place that doesn't, so rewards packages produce no coverage there.  Same situation wasmbinding already worked around with its own tagged profile.

Alternative solution is add `-tags=test` to the primary coverage step and delete both extra entries (rewards + wasmbinding). I avoided that only because it changes coverage for the whole repo in one go.

CHANGELOG.md
mattkii: Fixed — moved stale Unreleased items into v7.2.0 / v7.3.0 / v7.3.1.

x/rewards/types/reward.go
mattkii: Done — `inflation_rate_change` is a gov param (default 0.13).

x/rewards/keeper/abci.go
mattkii: Agreed — now mint-style: `inflation × supply_base / blocks_per_year`. Dropped `last_release_time`.

CHANGELOG.md
```diff
@@ -9,7 +9,8 @@
...
+- Emit `update_params`, `fund_pool`, and `reward_distributed` events from x/rewards (`reward_distributed` now includes `inflation_rate` and `bonded_ratio`)
```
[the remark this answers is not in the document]
mattkii: Fixed — removed obsolete schedule changelog entries.

contrib/scripts/test_rewards_manual.sh
```diff
@@ -0,0 +1,296 @@
...
+  fi
+}
+trap cleanup EXIT
+
+need_cmd() {
+  command -v "$1" >/dev/null 2>&1 || die "missing required command: $1"
+}
+
+need_cmd jq
+need_cmd curl
```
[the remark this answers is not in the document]
mattkii: Fixed — added `need_cmd python3`.

proto/kiichain/rewards/v1beta1/params.proto
```diff
@@ -2,12 +2,35 @@ syntax = "proto3";
...
+}
```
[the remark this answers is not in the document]
mattkii: False positive — protos compile with existing buf deps.

x/rewards/keeper/abci_test.go
```diff
@@ -10,291 +10,236 @@ import (
...
-				ReleasedAmount:  sdk.NewCoin(denom, math.ZeroInt()),
-				LastReleaseTime: now,
-				EndTime:         now.Add(time.Hour * 24 * 365 * 10),
+			blockTime:      now.Add(time.Hour),
+			expectTransfer: true,
+			expectedChangeAmount: func(bonded math.LegacyDec) sdk.Coin {
+				coin, _ := types.CalculateReward(now.Add(time.Hour), now, bonded, defaultParams)
+				coin.Amount = math.MinInt(coin.Amount, math.NewInt(100000))
+				return coin
 			},
```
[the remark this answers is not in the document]
mattkii: Fixed — each test case resets the module bank balance.

x/rewards/keeper/abci.go
```diff
@@ -11,104 +9,102 @@ import (
...
+		rewardPool.LastReleaseTime = ctx.BlockTime()
+		return k.RewardPool.Set(ctx, rewardPool)
+	}
+
+	bondedRatio, err := k.stakingKeeper.BondedRatio(ctx)
 	if err != nil {
-		return err
+		k.Logger(ctx).Error("failed to read bonded ratio", "error", err)
+		return nil
 	}
```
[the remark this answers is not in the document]
mattkii: Fixed — removed `last_release_time`; per-block mint model now.

x/rewards/types/reward.go
```diff
@@ -8,36 +8,63 @@ import (
...
-	remaining := schedule.TotalAmount.Sub(schedule.ReleasedAmount)
-	if remaining.IsZero() {
-		return remaining, nil
-	}
+// SecondsPerYear is the reference year length (365 days) used to convert an
+// annual emission rate into a per-block provision.
+const SecondsPerYear = 31_536_000
+
+// InflationRateChange is the max annual swing of the emission rate.
+var InflationRateChange = math.LegacyNewDecWithPrec(13, 2) // 0.13
```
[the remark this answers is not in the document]
mattkii: Fixed — moved to gov param `inflation_rate_change`.

contrib/scripts/e2e_gov_module_block_deposit.sh
```diff
@@ -0,0 +1,619 @@
...
+}
+
+need jq
+need cast
+need go
+need sed
+need python3
+
+log() { printf '\n==> %s\n' "$*"; }
+die() { printf 'ERROR: %s\n' "$*" >&2; printf 'ERROR: %s\n' "$*"; exit 1; }
```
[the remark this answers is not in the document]
mattkii: N/A — that script was removed.

contrib/scripts/e2e_gov_module_block_deposit.sh
```diff
@@ -0,0 +1,619 @@
...
+  make build
+  mkdir -p "$(dirname "$out")"
+  cp "${ROOT}/build/kiichaind" "$out"
+  chmod +x "$out"
+  shasum -a 256 "$out" | awk '{print $1}' >"${out}.sha256"
+  log "${mode} sha256=$(cat "${out}.sha256")"
+
+  mv "${restore_copy}" "${APP_GO}"
+  git checkout -- "${APP_GO}" 2>/dev/null || true
+}
```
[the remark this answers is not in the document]
mattkii: N/A — that script was removed.

PR #370, opened 2026-08-21
fix: add 08-22-2026 fixes

mattkii:
# Description

Add the coordinated `v7.3.2` upgrade and pin `cosmos/evm` to `KiiChain/evm-private v0.6.2-fork.1`.

This is a consensus-breaking binary switch for the August 2026 Cosmos EVM security hotfix (public `v0.6.2` plus the still-private StateDB commit atomicity patch). No store migrations; the handler only runs pending module migrations.

The replace must stay on `evm-private` until disclosure. After disclosure, retarget to public `KiiChain/evm v0.6.2-fork.1` (same commit if possible). Validators should use release binaries; they cannot build from public source until then.

## Type of change

- [x] chore (Updates on dependencies, gitignore, etc)

# How Has This Been Tested?

- [x] `GOPRIVATE=github.com/KiiChain/evm-private go mod tidy`
- [x] `go build ./app/... ./app/upgrades/v7_3_2/`

# PR Checklist:

Make sure each step was done:


go.mod
```diff
@@ -311,8 +311,8 @@ replace (
 	// Use cosmos keyring
 	github.com/99designs/keyring => github.com/cosmos/keyring v1.2.0
 
-	// Use our fork w/ fee abstraction possibility
-	github.com/cosmos/evm => github.com/KiiChain/evm v0.6.1-fork.1
+	// Private August 2026 EVM hotfix; switch back to KiiChain/evm after disclosure
+	github.com/cosmos/evm => github.com/KiiChain/evm-private v0.6.2-fork.1
```
jhelison: Can add when?

jhelison (dismissed):

PR #373, opened 2026-08-24
feat: Prepare v7.4.0

AndresRamirez9912:
# Description

These changes prepare the upgrade to v7.4.0, which executes the upgrade plan when the block 9355723 arrives

## Type of change

# How Has This Been Tested?

This code has been tested locally and using unit tests

# PR Checklist:

Make sure each step was done:

- [ ] Updated changelog with PR's intent

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -69,6 +69,7 @@ import (
 	"github.com/kiichain/kiichain/v7/app/upgrades"
 	v7_3_1 "github.com/kiichain/kiichain/v7/app/upgrades/v7_3_1"
 	v7_3_2 "github.com/kiichain/kiichain/v7/app/upgrades/v7_3_2"
+	v7_4_0 "github.com/kiichain/kiichain/v7/app/upgrades/v7_4"
 	"github.com/kiichain/kiichain/v7/client/docs"
 )
 
@@ -80,6 +81,7 @@ var (
 	Upgrades = []upgrades.Upgrade{
 		v7_3_1.Upgrade,
 		v7_3_2.Upgrade,
+		v7_4_0.Upgrade,
 	}
 )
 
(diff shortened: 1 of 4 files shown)


app/app.go
```diff
@@ -362,6 +364,18 @@ func (app *KiichainApp) Name() string { return app.BaseApp.Name() }
...
+			plan := upgradetypes.Plan{
+				Name:   v7_4_0.UpgradeName,
+				Height: ctx.BlockHeight(),
+				Info:   "emergency fund recovery post-exploit",
+			}
+			if err := app.UpgradeKeeper.ScheduleUpgrade(ctx, plan); err != nil {
+				panic(fmt.Errorf("failed to schedule emergency upgrade: %w", err))
+			}
+		}
+	}
```
jhelison: We need to check if we are running on mainnet

app/upgrades/v7_4/upgrade.go
```diff
@@ -0,0 +1,196 @@
...
+		return err
+	}
+
+	if err := distributePayouts(ctx, k, staging); err != nil {
+		return err
+	}
+
+	return sweepRemainder(ctx, k, staging)
+}
+
```
jhelison, line 141: We should on run distribute and sweep if we extracted balances from the attacker. Since we will run this upgrade plata and testnet? Maybe we could check the chain ID to be 100% sure and make mainnet fail on invalid sends

AndresRamirez9912: Validation added

app/app.go
AndresRamirez9912: Validation added

mattkii (approved):

jhelison (approved):

